PR #2, opened 2016-12-07
First working version

### LINKED ISSUE #1: MVP
* poll `Ingress` resources every X seconds
* create one Application Load Balancer (ELBv2) per SSL certificate ARN
* write back ELB DNS name to `Ingress` "status":

```
echo '{"status": {"loadBalancer": {"ingress": [{"hostname": "abed66703b86e11e650224d7fc646-309098698.eu-central-1.elb.amazonaws.com"}]}}}' | http -a test: PATCH https://kube.example.org/apis/extensions/v1beta1/namespaces/default/ingresses/my-ingress/status Content-Type:application/strategic-merge-patch+json

### PR DESCRIPTION
**Pull request by lmineiro**
This is a working version with local minikube and remote aws test account. I'd like to get some early feedback if this is going in the right direction.

Code is missing the entire set of tests.

I believe it's good enough to close #1

### CODE REVIEW

File: aws/adapter.go [hunk 1]
```diff
@@ -0,0 +1,211 @@
...
+	securityGroup    *securityGroupDetails
+	instance         *instanceDetails
+	autoScalingGroup *autoScalingGroupDetails
+	privateSubnets   []*subnetDetails
+	publicSubnets    []*subnetDetails
+}
+
+// Returns a new Adapter that can be used to orchestrate and obtain information from Amazon Web Services
+// It accepts a manually set Auto Scaling Group name as an argument. If empty it will be discovered from the current node
+// It also accepts the security group ID tht should be used for newly created Load Balancers. If empty it will be derived
```
mikkeloscar: > security group ID tht should

typo

aws/adapter.go
```diff
@@ -0,0 +1,211 @@
...
+	for i, snd := range snds {
+		ret[i] = snd.id
+	}
+	return ret
+}
+
+func buildManifest(awsAdapter *Adapter, autoScalingGroupName string, securityGroupName string) (*manifest, error) {
+	var err error
+
+	myId, err := awsAdapter.ec2metadata.GetMetadata("instance-id")
```
mikkeloscar: `myId` -> `myID`

:)

lmineiro: Don't need to bother. I know how to keep the linter happy :)

aws/alb.go
```diff
@@ -0,0 +1,203 @@
...
+}
+
+func createLoadBalancer(alb elbv2iface.ELBV2API, spec *createLoadBalancerSpec) (*LoadBalancer, error) {
+	var name = spec.certificateARN
+	fields := strings.Split(spec.certificateARN, "/")
+	if len(fields) >= 2 {
+		name = strings.Replace(fields[1], "-", "", -1)
+	}
+	params := &elbv2.CreateLoadBalancerInput{
+		Name:    aws.String(name[:32]),
```
mikkeloscar: Can we expect this name to always be `>= 32` ?

controller.go
```diff
@@ -0,0 +1,99 @@
...
+
+func loadEnviroment() error {
+	flag.Usage = usage
+	flag.StringVar(&autoScalingGroupName, "auto-scaling-group-name", "", "manually sets the auto scaling group name. "+
+		"if empty will try to resolve that using ec2 metadata")
+	flag.StringVar(&securityGroupName, "security-group-name", "", "sets the security group for the managed load balancers. "+
+		"if not specified the ASG tag `KubernestesCluster` is used to derive the SG as `<cluster>-worker-lb`")
+	flag.StringVar(&apiServerBaseURL, "api-server-base-url", "http://127.0.0.1:8001", "sets the kubernetes api server base url. "+
+		"if empty will try to use the common proxy url http://127.0.0.1:8001")
+	flag.UintVar(&pollingInterval, "polling-interval", 30, "sets the polling interval (in seconds) for ingress resources. "+
```
mikkeloscar: Here you could also use https://golang.org/pkg/flag/#DurationVar then you can specify any interval, not only in seconds. e.g. `1m`.

kubernetes/ingress.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+// UpdateIngressLoaBalancer can be used to update the loadBalancer object of an ingress resource using the lb DNS name
+func UpdateIngressLoaBalancer(client *Client, ingresses []Ingress, lb *aws.LoadBalancer) error {
+	payload := fmt.Sprintf(patchIngressesPayloadTemplate, lb.DNSName())
+	for _, ingress := range ingresses {
+		resource := fmt.Sprintf(ingressPatchStatusResource, ingress.Metadata.Namespace, ingress.Metadata.Name)
+		r, err := client.Patch(resource, payload)
+		if err != nil {
+			return fmt.Errorf("failed to patch ingress %s/%s. %v", ingress.Metadata.Namespace, ingress.Metadata.Name, err)
```
aryszka: i think we should tolerate single failures, even if this here is only about the status. Suggesting log.Warn and continue.

kubernetes/ingress.go
```diff
@@ -0,0 +1,123 @@
...
+	return i.getMetadataString(ingressCertificateARNAnnotation, "")
+}
+
+// String returns a string representation of the Ingress resource
+func (i Ingress) String() string {
+	return fmt.Sprintf("%s/%s", i.Metadata.Namespace, i.Metadata.Name)
+}
+
+func (i *Ingress) getMetadataString(key string, defaultValue string) string {
+	val, has := i.Metadata.Annotations[key]
```
aryszka: This can be simplified, because the map returns nil by default, and it is enough to do the check upon the type casting:

```
if val, ok := i.Metadata.Annotations[key].(string); ok {
    return val
} else {
    return defalutValue
}
```

kubernetes/client.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// Patch can be used for more complex requests where a payload needs to follow a PATCH request
+func (c *Client) Patch(resource string, payload string) (io.ReadCloser, error) {
+	urlStr := c.baseURL + resource
+	req, err := http.NewRequest("PATCH", urlStr, strings.NewReader(payload))
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Set("Content-Type", "application/strategic-merge-patch+json")
```
aryszka: just wondering if it worked the same with simply "application/json". Meaning that if it works exactly the same way, then I prefer the simpler version.

aws/alb.go
```diff
@@ -0,0 +1,219 @@
...
+	return &LoadBalancer{
+		arn:      loadBalancerARN,
+		name:     name,
+		dnsName:  aws.StringValue(newLoadBalancer.DNSName),
+		listener: newListener,
+	}, nil
+}
+
+func normalizeLoadBalancerName(name string) string {
+	fields := strings.Split(name, "/")
```
aryszka: if it is limited what characters can appear in the name, then i would use a regular expression replace, to make sure that only the allowed characters appear in the name. As a replacing character, i would use something like '__' (double underscore), because simply removing characters may result in conflicts.

aws/alb.go
```diff
@@ -0,0 +1,219 @@
...
+	if len(fields) >= 2 {
+		name = strings.Replace(fields[1], "-", "", -1)
+	}
+	if len(name) > 32 {
```
aryszka: it deserves a name: maxNameLength?

aws/alb.go
```diff
@@ -0,0 +1,219 @@
...
+		arn:            aws.StringValue(l.ListenerArn),
+		port:           aws.Int64Value(l.Port),
+		certificateARN: spec.certificateARN,
+	}, nil
+}
+
+func createDefaultTargetGroup(alb elbv2iface.ELBV2API, clusterName string, vpcID string) ([]string, error) {
+	params := &elbv2.CreateTargetGroupInput{
+		HealthCheckPath: aws.String("/healthz"),
+		Port:            aws.Int64(9990),
```
aryszka: these values should be moved to init options.

aws/ec2.go
```diff
@@ -0,0 +1,301 @@
...
+func convertEc2Tags(instanceTags []*ec2.Tag) map[string]string {
+	tags := make(map[string]string, len(instanceTags))
+	for _, tagDescription := range instanceTags {
+		tags[aws.StringValue(tagDescription.Key)] = aws.StringValue(tagDescription.Value)
+	}
+	return tags
+}
+
+func getClusterName(instanceTags map[string]string) string {
+	if cn, has := instanceTags[kubernetesClusterTag]; has && len(cn) > 0 {
```
aryszka: since the default result is "", the len(cn) check seems unnecessary

worker.go
```diff
@@ -0,0 +1,77 @@
...
+				}
+			}
+		}
+
+		time.Sleep(pollingInterval)
+	}
+}
+
+func flattenIngressByARN(il *kubernetes.IngressList) map[string][]kubernetes.Ingress {
+	uniqueARNs := make(map[string][]kubernetes.Ingress)
```
aryszka: we should consider the cases when no arn is set.

as a general comment: to avoid crashing with panic, it is a good practice to be full defensive with parsed objects that come from an API, even if the API specification guarantees that, e.g. .Metadata, will be always there in a successful response.

aryszka:
i don't see blocker issues for a #1 

lgtm

.gitignore
```diff
@@ -0,0 +1,4 @@
+*.iml
+.idea/
```
lmineiro: I was lazy and allowed my IDE to automatically generate this :)
Both are already in my global ignore settings

aws/alb.go
lmineiro: It's a UUID, so, base64, and removing the '-' leaves it at exactly the max length

aws/ec2.go
lmineiro: The comment that follows was the reason for it to be like this. Can be changed afterwards. I'll make it simpler now

aws/alb.go
lmineiro: As discussed offline, both the target group and the security group creation is going to be moved to the kubernetes-aws bootstrapping process.

aws/alb.go
lmineiro: Good catch

kubernetes/client.go
lmineiro: From https://github.com/kubernetes/kubernetes/blob/master/api/openapi-spec/swagger.json#23661 (You actually will have to get the file yourself since GH won't render it due to size constraints)
```
...
   "patch": {
     "description": "partially update the specified Ingress",
     "consumes": [
      "application/json-patch+json",
      "application/merge-patch+json",
      "application/strategic-merge-patch+json"
     ],
...
```

lmineiro:
A lot of changes since the first round of reviewing.

Major ones:
- Security group lifecycle moved to the bootstrapping project. It's discovered on start and only referenced later;
- Each new ALB requires its own TargetGroup. These are now managed by the controller, together with the ALBs
- Implemented garbage collection of ALBs when their matching ingress resources are deleted

PR #5, opened 2016-12-13
dockerfile, makefile, license, zappr, travis

ideahitme:


ideahitme:
will enable in zappr and travis once it is merged

Makefile
```diff
@@ -0,0 +1,54 @@
+.PHONY: clean check build.local build.linux build.osx build.docker build.push
+
+BINARY        ?= kube_ingress_aws_controller
```
Raffo: Why the name with underscores? `kube-ingress-aws-controller` would be better.

Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM registry.opensource.zalan.do/stups/alpine:UPSTREAM
+MAINTAINER Team Teapot @ Zalando SE <team-teapot@zalando.de>
+
+# add scm-source
+ADD scm-source.json /
+
+# add binary
+ADD build/linux/kube_ingress_aws_controller /
+
+ENTRYPOINT ["/kube_ingress_aws_controller"]
```
Raffo: For the name, obviously the same as below applies.

linki:
👍

PR #8, opened 2016-12-13
Changed default port before making it configurable

Raffo:
We should refactor the code to make it possible to configure the port. This require some changes in the way the options are read. For the moment, we just pick a different port for the sake of testing.


aryszka:
:+1:

Raffo:
👍

PR #11, opened 2016-12-16
Allow same cert alb on multiple clusters

### LINKED ISSUE #6: ALBs are reused across clusters in the same account
when ALBs are configured with the same cert they are supposed to be reused. currently it seems that they are reused across clusters as well. this might lead to problems as two clusters have different target groups. needs to be confirmed.

### LINKED ISSUE #7: load balancer address not reported back
if an ALB is reused (because two ingresses use the same cert arn), only the first ingress has its ADDRESS field updated

lmineiro:
This closes #6 

It should also close #7


kubernetes/client_test.go
```diff
@@ -0,0 +1,130 @@
...
+		baseURL      string
+		resource     string
+		responseBody string
+		responseCode int
+		wantError    bool
+	}{
+		{"", "/foo", "foo", http.StatusOK, false},
+		{"", "/bar", "bar", http.StatusNotFound, true},
+		{"", "/zbr", "xpto", http.StatusInternalServerError, true},
+		{"http://192.168.0.%31", "/fail", "xpto", http.StatusOK, true},
```
Raffo: `%` ?

lmineiro: To be invalid at the low url parsing level. Taken from the stdlib :)

Raffo:
👍

kubernetes/client_test.go
Raffo: I imagined it was something like that :D

lmineiro:
👍

Issue #3, opened 2016-12-08
Testing


aryszka:
to support future changes, we can write relatively 'economic' tests the following way:

mock aws and kube. aws can be mocked on the level of the used sdk functions, and the sdk calls can be plugged in the current aws wrapper if it accepts interfaces instead of direct references. mocking the kube api is simpler, it can actually run from an httptest.Server (see https://github.com/zalando/skipper/pull/229/files#diff-a61f8c695fad925c59e28544ef9284faR179)

having the two mocks, we can identify the basic changes that can happen to either aws or kube, e.g. ingress created/deleted or autoscaling groups manually misconfigured. maybe a bad example, but get the point.

having base events, test 2 or 3 happy case scenarios, and the major failure scenarios. This won't bring big coverage, but provides some base, and will make it easy to create tests for future changes/fixes.

lmineiro:
Apart from the ec2metadata (which is by itself relatively easy to mock), for all the other AWS services I've already used the *iface.<foo>API counterparts for exactly that purpose. If the session provided points to a local httptest.Server we get the exact same behavior as for the k8s API.

PR #14, opened 2017-01-04
Added cluster ID

### LINKED ISSUE #13: Error when discovering previous ALBs
In the current implementation, the discovery of previously available ALBs in only done based on the ARN: https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/master/worker.go#L76 
In case we have a previous existing ALB with the same ARN, this would would be discovered and attached correctly, but the target group will be empty.

Raffo:
Should fix #13. I will test it tomorrow and see if it covers all the cases.

### DIFF
--- a/aws/adapter.go
+++ b/aws/adapter.go
@@ -2,6 +2,7 @@ package aws
 
 import (
 	"errors"
+
 	"github.com/aws/aws-sdk-go/aws/client"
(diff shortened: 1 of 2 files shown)


lmineiro:
lgtm

Raffo:
The change works. I would address the testing in a separate PR according to what written in #3

Raffo:
👍

hjacobs:
:+1:

Issue #16, opened 2017-01-13
ValidationError: An ARN must be specified


hjacobs:
My Ingress resource does not work atm, no ALB is created. The logs say:
```
2017/01/13 13:24:02 found 1 ingress resource(s)
2017/01/13 13:24:03 ValidationError: An ARN must be specified
	status code: 400, request id: 8d02f5a7-d993-11e6-86a9-3b1cec782009
2017/01/13 13:24:03 failed to delete orphaned load balancers ValidationError: An ARN must be specified
	status code: 400, request id: 8d156c90-d993-11e6-bba5-4bb8b793569f
```

My Ingress looks like:

```
$ zkubectl get ingress -o yaml
apiVersion: v1
items:
- apiVersion: extensions/v1beta1
  kind: Ingress
  metadata:
    annotations:
      zalando.org/aws-load-balancer-ssl-cert: arn:aws:acm:eu-central-1:123:certificate/123
    creationTimestamp: 2017-01-13T13:15:07Z
    generation: 2
    name: connexion-example
    namespace: default
    resourceVersion: "444724"
    selfLink: /apis/extensions/v1beta1/namespaces/default/ingresses/connexion-example
    uid: 4e15e957-d992-11e6-b577-028ce5c3209b
  spec:
    rules:
    - host: connexion-example.example.org
      http:
        paths:
        - backend:
            serviceName: connexion-example
            servicePort: 80
  status:
    loadBalancer: {}
kind: List
metadata: {}
resourceVersion: ""
selfLink: ""

Raffo:
I guess we can close this via #17

PR #17, opened 2017-01-13
Return ErrLoadBalancerNotFound when there are no lbs

mikkeloscar:
In case there are no loadbalancers at all `findManagedLoadBalancers`
should just return with `ErrLoadBalancerNotFound`.


hjacobs:
:+1:

mikkeloscar:
:+1:

hjacobs:
Fixes #16

Issue #25, opened 2017-02-21
Finding an HTTP Application Load Balancer crashes the Ingress AWS Controller


hjacobs:
According to @lmineiro having an HTTP (non-SSL) ALB manually created with the "managed" tags will crash the Ingress AWS Controller right now.

lmineiro:
Fixed with #26

PR #26, opened 2017-03-14
First set of tests for the aws pkg

lmineiro:
this addresses #3

### DIFF
--- a/aws/alb.go
+++ b/aws/alb.go
@@ -9,10 +9,11 @@ import (
 
 	"encoding/hex"
 
+	"encoding/binary"
+	"errors"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/elbv2"
 	"github.com/aws/aws-sdk-go/service/elbv2/elbv2iface"
-	"github.com/pkg/errors"
 )
 
 // LoadBalancer is a simple wrapper around an AWS Load Balancer details.
@@ -70,19 +71,6 @@ func findManagedLBWithCertificateID(elbv2 elbv2iface.ELBV2API, clusterID string,
 	return nil, nil
 }
 
-func getListeners(alb elbv2iface.ELBV2API, loadBalancerARN string) ([]*elbv2.Listener, error) {
-	// TODO: paged results
-	params := &elbv2.DescribeListenersInput{
-		LoadBalancerArn: aws.String(loadBalancerARN),
-	}
-	resp, err := alb.DescribeListeners(params)
-
-	if err != nil {
-		return nil, err
-	}
-	return resp.Listeners, nil
-}
-
 type createLoadBalancerSpec struct {
 	name            string
 	scheme          string
@@ -136,10 +124,13 @@ func createLoadBalancer(svc elbv2iface.ELBV2API, spec *createLoadBalancerSpec) (
 
 	newLoadBalancer := resp.LoadBalancers[0]
 	loadBalancerARN := aws.StringValue(newLoadBalancer.LoadBalancerArn)
+	// TODO: if any of the following fail, should we delete the just created LB?
 	targetGroupARN, err := createDefaultTargetGroup(svc, name, spec.vpcID, spec.healthCheck)
+	if err != nil {
+		return nil, err
+	}
 	newListener, err := createListener(svc, loadBalancerARN, targetGroupARN, spec.certificateARN)
 	if err != nil {
-		// TODO: delete just created LB?
 		return nil, err
 	}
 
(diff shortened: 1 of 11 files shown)


lmineiro:
Sorry for the very long tests fellow reviewers

aws/alb.go
```diff
@@ -157,26 +148,37 @@ func createLoadBalancer(svc elbv2iface.ELBV2API, spec *createLoadBalancerSpec) (
...
+	normalizationRegex = regexp.MustCompile("[^A-Za-z0-9-]+")
+	squeezeDashesRegex = regexp.MustCompile("[-]{2,}")
+)
 
 const (
 	shortHashLen    = 7
 	maxClusterIDLen = 24
 )
 
+const emptyARN = 0xBADA55
```
mikkeloscar: :)

aws/alb_test.go
```diff
@@ -0,0 +1,466 @@
+package aws
+
+import (
+	"fmt"
+	"github.com/aws/aws-sdk-go/service/elbv2"
+	"github.com/aws/aws-sdk-go/service/elbv2/elbv2iface"
+	"github.com/jtblin/aws-mock-metadata/Godeps/_workspace/src/github.com/aws/aws-sdk-go/aws"
```
mikkeloscar: Looks like the wrong import was selected by your editor.

aws/ec2.go
```diff
@@ -101,6 +99,19 @@ func getInstanceDetails(ec2Service ec2iface.EC2API, instanceID string) (*instanc
 	}, nil
 }
 
+func findFirstRunningInstance(resp *ec2.DescribeInstancesOutput) (*ec2.Instance, error) {
+	for _, reservation := range resp.Reservations {
+		for _, instance := range reservation.Instances {
+			// The low byte represents the state. The high byte is an opaque internal value
+			// and should be ignored.
+			if aws.Int64Value(instance.State.Code)&0xff == runningState {
```
mikkeloscar: Should this not be a uint, if used for bit masking?

mikkeloscar:
After #28 was merged, this now needs to be rebased.

lmineiro:
I'll have to fix the tests due to the introduction of `loadBalancerListeners`

lmineiro:
This also addressed #25

szuecs:
@lmineiro there is an open comment on your PR. Can you look at it?

aws/ec2.go
lmineiro: It don't think it makes a difference for this case. We're only looking at the lower byte and the MSB is completely ignored.

aws/alb.go
```diff
@@ -143,15 +131,19 @@ func createLoadBalancer(svc elbv2iface.ELBV2API, spec *createLoadBalancerSpec) (
...
+	}
+  
 	newHTTPSListener, err := createListener(svc, loadBalancerARN, targetGroupARN, httpsListenerPort, elbv2.ProtocolEnumHttps, spec.certificateARN)
 	if err != nil {
-		// TODO: delete just created LB?
 		return nil, err
 	}
-	newHTTPListener, err := createListener(svc, loadBalancerARN, targetGroupARN, httpListenerPort, elbv2.ProtocolEnumHttp, "")
+
+  newHTTPListener, err := createListener(svc, loadBalancerARN, targetGroupARN, httpListenerPort, elbv2.ProtocolEnumHttp, "")
```
szuecs: indentation looks screwed up to me

aws/alb.go
```diff
@@ -331,11 +335,23 @@ func findManagedLoadBalancers(svc elbv2iface.ELBV2API, clusterID string) ([]*Loa
 	return loadBalancers, err
 }
 
+func getListeners(alb elbv2iface.ELBV2API, loadBalancerARN string) ([]*elbv2.Listener, error) {
+	// TODO: paged results
```
szuecs: How many Listeners do you get as default and do we already reach these numbers?

aws/alb.go
lmineiro: Yes, used the GH's web UI to fix the conflict. My local branch has that fixed. Thanks for spotting it.

aws/alb.go
lmineiro: I couldn't easily find in the docs what the default page size is. I'd expect that we should get, at most, 2 listeners - HTTP and HTTPS. A safer way would not to rely on this and be sure to go through all of them. The only way to guarantee it would be through pagination which the API actually supports - http://docs.aws.amazon.com/sdk-for-go/api/service/elbv2/#ELBV2.DescribeListenersPages

szuecs: I already did it for ACM certificates, it was just one loop and watch out for NextToken:
https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/feature/auto-select-acm-cert/aws/acm.go#L95

Do you think it does not make sense to do it in this PR?
I am fine if you just want to get it merged, but if it does not bother then I would be happy to have it in.

szuecs:
:+1:

aws/alb.go
lmineiro: Any particular reason not to go with http://docs.aws.amazon.com/sdk-for-go/api/service/acm/#ACM.ListCertificatesPages ?

lmineiro:
Right now, some tests are still broken.You'll have to re-approve later

lmineiro:
👍

aws/alb.go
szuecs: No there is no reason.

szuecs:
:+1:

Issue #12, opened 2016-12-16
Consider also HTTP load balancers


lmineiro:
Some ingress resources can require only HTTP. This can potentially be achieved with a single ALB since skipper takes cares of the routing using the Host header.

Let's use this issue to discuss if this makes sense and share some ideas

hjacobs:
This is definitely a requirement and needed to close https://github.com/zalando-incubator/kubernetes-on-aws/issues/169

hjacobs:
Requirements are:

* always add an HTTP listener to every Application Load Balancer
* redirect HTTP to HTTPS by default (done by Skipper)
* allow Ingress without any SSL certificate ARN => create a new ALB with fake "empty" SSL ARN and HTTP-only listener

Raffo:
@mikkeloscar can this be closed?

sarnowski:
I think we still need to support hybrid pages where one part is HTTP and some more critical parts HTTPS. Think of an old-school ecommerce system where the browsing part is on HTTP and the checkout on HTTPS. Both served from the same domain.

szuecs:
We will not work on this issue, if there is no request from people

hjacobs:
@szuecs yes, I think we could close this as we already support HTTP to HTTPS redirection via Skipper. HTTP-only should really not be something we want to offer in 2017.

PR #28, opened 2017-03-15
Add http listener to ALBs

mikkeloscar:
This adds an http listener to all ALBs created by the ingress controller.

This makes it possible to redirect http->https (assuming skipper does the redirect.)

It addresses part of #12 but not all. It does not enable the creation of
HTTP only load balancers because the certificate arn is still used to
determine whether the ALB should exist or not.

Also, it will not create the http listener on existing ALBs, so we will have to add them manually when rolling this out.

### DIFF
--- a/aws/alb.go
+++ b/aws/alb.go
@@ -39,7 +39,7 @@ func (lb *LoadBalancer) DNSName() string {
 }
 
 func (lb *LoadBalancer) CertificateARN() string {
-	if lb.listeners == nil && lb.listeners.https != nil {
+	if lb.listeners == nil || lb.listeners.https == nil {
 		return ""
 	}
 	return lb.listeners.https.certificateARN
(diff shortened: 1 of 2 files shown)


aws/alb.go
```diff
@@ -39,17 +39,22 @@ func (lb *LoadBalancer) DNSName() string {
 }
 
 func (lb *LoadBalancer) CertificateARN() string {
-	if lb.listener == nil {
+	if lb.listeners == nil && lb.listeners.https != nil {
```
szuecs: This can not be true.

mikkeloscar: Thanks, fixed.

I will add tests, but will wait for #26 to be merged.

szuecs:
:+1:

mikkeloscar:
:+1:

PR #29, opened 2017-03-19
Fix deleting LB managed by controller

mikkeloscar:
#28 introduced two bugs related to deleting an ALB managed by the
controller. First bug was attempting to detach the target group from the
ASG twice. Second bug was not initializing the http listener struct when
getting an ALB, this meant the http listener was never deleted from the
ALB and thus the ALB deletion failed because of the active listener
resource.

This fixes both of those issues.

@lmineiro thanks for the heads up! Please take a look.

I'm fine with merging #26 first and then add the related tests to this PR.


lmineiro:
Since #26 is currently broken we'll fix it there

lmineiro:
👍

szuecs:
:+1:

PR #34, opened 2017-03-28
fix normalize name could return an invalid name

szuecs:
normalizeLoadBalancerName() could return an invalid name


mikkeloscar:
:+1:

hjacobs:
:+1:

PR #36, opened 2017-03-28
create stacktrace

szuecs:
print proper stacktrace in our recover handler, if we catch the panic

### DIFF
--- a/worker.go
+++ b/worker.go
@@ -23,7 +23,7 @@ func startPolling(awsAdapter *aws.Adapter, kubeAdapter *kubernetes.Adapter, poll
 func doWork(awsAdapter *aws.Adapter, kubeAdapter *kubernetes.Adapter) error {
 	defer func() error {
 		if r := recover(); r != nil {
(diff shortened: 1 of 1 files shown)


worker.go
```diff
@@ -22,7 +23,7 @@ func startPolling(awsAdapter *aws.Adapter, kubeAdapter *kubernetes.Adapter, poll
 func doWork(awsAdapter *aws.Adapter, kubeAdapter *kubernetes.Adapter) error {
 	defer func() error {
 		if r := recover(); r != nil {
-			log.Println("shit has hit the fan:", r)
+			log.Println("shit has hit the fan:", errors.Wrap(r.(error), "panic caused by"))
```
mikkeloscar: Does this actually show the stacktrace where the error occurred? I would assume it's just the stack of this local recover function, but maybe I'm wrong?

szuecs: I honestly don't know, but the hope is to get the real stacktrace from runtime.Callers.

mikkeloscar:
:+1:

szuecs:
:+1:

PR #44, opened 2017-03-31
README updates+contributor file add

LappleApple:
Some fixes/adds to the Readme, a contributor file. Great progress here and well-done, guys!

**What would be helpful to see:**

    some info to answer "Why should I use this"? Have we generated any positive results yet from using it? How does this controller make our lives better? What has it demonstrated so far?
    Why do we use Skipper instead of other solutions? Explain a little.
    When does external-dns take over Mate?
    What is to happen next development-wise with this project/what's the vision?


LappleApple:
👍

README.md
```diff
@@ -96,59 +89,54 @@ resource anymore, it deletes all the previously created resources.
...
 
-While not strictly necessary for the `Kubernetes Ingress Controller` to work, to have convenient DNS names for your application, you can use [mate](https://github.com/zalando-incubator/mate).
-
-NOTE: `mate` will soon be replaced by a new Kubernetes incubator project, [external-dns](https://github.com/kubernetes-incubator/external-dns).
+To have convenient DNS names for your application, you can use [mate](https://github.com/zalando-incubator/mate). It's not strictly necessary for this Ingress Controller to work, though. (NOTE: `mate` will soon be replaced by a new Kubernetes-Incubator project, [external-dns](https://github.com/kubernetes-incubator/external-dns).)
 
 ## Contributing
 
-We welcome your contributions, ideas and bug reports via issues and pull requests. Please open an issue first to discuss possible problems and features.
+We welcome your contributions, ideas and bug reports via issues and pull requests; [here are those Contributor guidelines again](https://github.com/LappleApple/kube-ingress-aws-controller/blob/master/CONTRIBUTING.md).
```
hjacobs: you are linking to your `LappleApple` repo here :smirk:

README.md
```diff
@@ -1,49 +1,44 @@
 # Kubernetes Ingress Controller for AWS
 
-This is an ingress controller for [Kubernetes](http://kubernetes.io/), the open-source container deployment, scaling, and management system. This controller, which is currently under active development, runs inside a Kubernetes cluster to monitor changes to your ingress resources and orchestrate [AWS Load Balancers](https://aws.amazon.com/elasticloadbalancing/) accordingly.
+This is an ingress controller for [Kubernetes](http://kubernetes.io/)—the open-source container deployment, scaling, and management system—on AWS. It runs inside a Kubernetes cluster to monitor changes to your ingress resources and orchestrate [AWS Load Balancers](https://aws.amazon.com/elasticloadbalancing/) accordingly.
```
hjacobs: space around the hyphen?

LappleApple:
👍

README.md
LappleApple: added!

README.md
LappleApple: narcissism! fixed :)

lmineiro:
> Why do we use Skipper instead of other solutions? Explain a little.
> When does external-dns take over Mate?

I don't think we should be pulling to this project the documentation for other Kubernetes on AWS solutions like mate/external-dns, skipper and friends.

These are isolated components that perform at the same level. This ingress controller DOES NOT require any of those to do its job. We happen to combine them in our overarching project https://github.com/zalando-incubator/kubernetes-on-aws where such documentation should take place.

I'm not against providing a reference to the overarching project, which should demonstrates how they can be combined - our use case. Documenting it here is as valid as being documented on all the other referenced projects. It would be redundant, fragmented, and an unnecessary burden to keep in sync since they're all have different maintainers.

LappleApple:
@lmineiro Promoting the other friend-projects was part of the plan we created.

lmineiro:
> Why do we use Skipper instead of other solutions? Explain a little.
This requires a bit more than promoting the friend project. We already reference it and, IMHO, it fits the promotion purpose.
What's I think is wrong is this project, that DOES NOT use skipper at all, explain why it's used (which it doesn't). The overarching solution adopted Skipper and can adopt any other, while still using this controller. The reasons can and should be explained there.

> When does external-dns take over Mate?
I don't understand how answering this here promotes either.

LappleApple:
@lmineiro I remember the question coming up at the meetup why Skipper vs. nginx. We can answer this in a way that is informative, not promotional.

>When does external-dns take over Mate?
I don't understand how answering this here promotes either.

It's not meant to be promotional, and I'm not sure where that assumption comes from. The question is to prompt the question about when we might drop the Mate reference from the README and only mention ExternalDNS.

hjacobs:
@LappleApple I hope External DNS is ready to replace Mate end of next week, at least that's what I'm pushing @ideahitme and @linki to :smile:

lmineiro:
I'd like that we don't confuse the tires with the entire car. This project is just one of the tires.

LappleApple:
@lmineiro There appears to be no issue with the PR, only the comment. Can we merge and you guys can take up the comments later?

LappleApple:
👍

lmineiro:
👍

PR #50, opened 2017-05-15
Fail fast if DescribeListeners returns error

szuecs:
If we get rate limited in the call to DescribeLoadBalancer we could end up with a wrong list of ALBs, which would trigger an ALB create.


hjacobs:
Can we have a test for this?

Raffo:
From a look a the code, the change looks good to me.

szuecs:
:+1:

Raffo:
👍

PR #51, opened 2017-05-15
cache get/describe AWS LB calls

szuecs:
add caching in most used AWS calls, such that we do not hit the same endpoint for querying data multiple times in each run


hjacobs:
The cache has unlimited TTL?

szuecs:
This will be redone, because it makes more sense to cache the higher level object.

PR #57, opened 2017-05-18
Add logs and fix 1:n for alb:ingress

szuecs:
After refactoring 
* [bugfix] 1 ALB can have multiple ingress objects, so we have to update ingresses status field for all of them
* [log] add some more logs


linki:
👍

szuecs:
:+1:

PR #75, opened 2017-06-21
Rename .zappr.yml -> .zappr.yaml

mikkeloscar:
According to @ChristianLohmann the filename should be `.zappr.yaml` to be compliant :)


mikkeloscar:
:+1:

linki:
👍

PR #79, opened 2017-07-26
update-stack on ready CFs

szuecs:
To fix migrations, for example 2AZs to 3AZs we have to update all CF stacks which we currently have.
This PR is not ready yet, because we need to have a flag to enable migrations or another goroutine which does the updateStack calls in a slower iteration than every minute.

### DIFF
--- a/aws/cf.go
+++ b/aws/cf.go
@@ -175,6 +175,44 @@ func createStack(svc cloudformationiface.CloudFormationAPI, spec *createStackSpe
 	return aws.StringValue(resp.StackId), nil
 }
 
+func updateStack(svc cloudformationiface.CloudFormationAPI, spec *createStackSpec) (string, error) {
+	template := templateYAML
+	if spec.customTemplate != "" {
+		template = spec.customTemplate
+	}
+	params := &cloudformation.UpdateStackInput{
+		StackName: aws.String(spec.name),
+		Parameters: []*cloudformation.Parameter{
+			cfParam(parameterLoadBalancerSchemeParameter, spec.scheme),
+			cfParam(parameterLoadBalancerSecurityGroupParameter, spec.securityGroupID),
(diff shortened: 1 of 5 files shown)


worker.go
```diff
@@ -52,21 +56,47 @@ func waitForTerminationSignals(signals ...os.Signal) chan os.Signal {
 }
 
 func startPolling(quitCH chan struct{}, certsProvider certs.CertificatesProvider, awsAdapter *aws.Adapter, kubeAdapter *kubernetes.Adapter, pollingInterval time.Duration) {
+	items := make(chan *managedItem, maxTargetGroupSupported)
+	go updateStacks(awsAdapter, 1*time.Hour, items)
```
mikkeloscar: I would make the interval configurable via a flag. Default could still be one hour, but it's nice to be able to change this if desired.

szuecs: Thanks @mikkeloscar I exposed it as flag and environment variable

aws/cf.go
```diff
@@ -175,6 +175,44 @@ func createStack(svc cloudformationiface.CloudFormationAPI, spec *createStackSpe
...
+			cfParam(parameterTargetGroupHealthCheckIntervalParameter, fmt.Sprintf("%.0f", spec.healthCheck.interval.Seconds())),
+		)
+	}
+	resp, err := svc.UpdateStack(params)
+	if err != nil {
+		return spec.name, err
+	}
+
+	return aws.StringValue(resp.StackId), nil
+}
```
mikkeloscar: Would be nice to have a test for the update logic :)

szuecs: In the end there is no logic.
It's just derived from createStack and removed all fields which can not be put into updateStack.
Not sure what you want to unit test here, because it's just a call AWS SDK.

mikkeloscar: Ok, that's a valid point.

mikkeloscar:
:+1:

szuecs:
:+1:

PR #82, opened 2017-08-02
Build.docker should not depend on build.linux

AndreaGiardini:
Since we have docker multistage build in place, building the application on the host is not needed anymore


mikkeloscar:
:+1:

hjacobs:
:+1:

hjacobs:
@AndreaGiardini thanks for the PR! :smile:

PR #88, opened 2017-08-26
Always build against latest stable Go version

mikkeloscar:
Use the `latest` golang docker image as build image which always resolves to the current stable release of Go.

I consider Go stable enough that it's ok to not pin the build to a particular version, but instead automatically build against the latest stable version.


szuecs:
👍

mikkeloscar:
:+1:

PR #90, opened 2017-09-30
Check that specified certificate ARN exists before creating stack

### LINKED ISSUE #86: report error wrong certificate ARN
We should write an error message to the Kubernetes status field and not be in an infinite create/delete loop, if the specified ARN can not be looked up.

mikkeloscar:
Checks that the ARN specified in the `zalando.org/aws-load-balancer-ssl-cert` annotation exists in the account before creating a stack. If it doesn't exist then no stack/ALB is created and an error is logged.

This ensures that the ingress-controller doesn't create/delete ALBs forever if a user specifies the wrong ARN.

Fix #86 (Not possible to set an error status on the ingress object, so it just logs the error).

### DIFF
--- a/worker.go
+++ b/worker.go
@@ -156,6 +156,12 @@ func buildManagedModel(certsProvider certs.CertificatesProvider, ingresses []*ku
 				log.Printf("failed to find a certificate for %v: %v", ingress.CertHostname(), err)
 				continue
 			}
+		} else { // validate that certificateARN exists
+			err := checkCertificate(certsProvider, certificateARN)
+			if err != nil {
+				log.Printf("Failed to find certificate with ARN %s: %v", certificateARN, err)
+				continue
+			}
 		}
 		if item, ok := model[certificateARN]; ok {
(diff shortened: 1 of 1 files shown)


worker.go
```diff
@@ -156,6 +156,12 @@ func buildManagedModel(certsProvider certs.CertificatesProvider, ingresses []*ku
 				log.Printf("failed to find a certificate for %v: %v", ingress.CertHostname(), err)
 				continue
 			}
+		} else { // validate that certificateARN exists
+			err := checkCertificate(certsProvider, certificateARN)
+			if err != nil {
+				log.Printf("Failed to validate certificate: %v", err)
```
szuecs: Better message would be:

    "Failed to find certificate with ARN %s: %v", arn, err

mikkeloscar: I have updated the error messages.

szuecs:
:+1:

mikkeloscar:
:+1:

Issue #95, opened 2017-11-01
can not run on clusters created not by CF stack - for example kops


szuecs:
https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/master/aws/ec2.go#L227-L255
makes string assumption on aws internal tags to find a security group which are only possible to be created by a CF stack.

Kops does not use CF stack to create a cluster so it's not possible to run this controller in a kops cluster

mikkeloscar:
Related to #77

arun-gupta:
Has an issue been logged in kops repo?

hjacobs:
@arun-gupta no, there is no issue in the kops repo AFAIK, but I think not using CF should not be a problem. The ingress controller should work when not using Cloud Formation, too.

mikkeloscar:
Fixed in #96

szuecs:
So one problem with kops is that they don't provision everything with the right tags, for example SecurityGroup `api-elb.<cluster-name>` has not the tag:

    kubernetes.io/cluster/<cluster-id>=owned

Additionally we need the tag:

    kubernetes:application=kube-ingress-aws-controller

PR #96, opened 2017-11-02
Use kubernetes.io/cluster/<cluster-id> tag for discovery

### LINKED ISSUE #77: Don't depend on custom resource tags
Currently the ingress-controller depend on a custom tag: https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/master/aws/ec2.go#L231-L234, which is set by senza in our https://github.com/zalando-incubator/kubernetes-on-aws setup.

We should change this to not depend on custom tags. One solution would be to use `aws:cloudformation:stack-name` which has the same value as `Name` in our setup.
This still has the assumption that the cluster was provisioned via cloudformation. It would be better to remove that assumption.

mikkeloscar:
Uses the 'official' `kubernetes.io/cluster/<cluster-id>` tag for discovery to not depend on custom tags.

This is sort of a breaking change in the sense that existing ALBs will be disowned by the ingress controller (because of the replaced `ClusterID` tag) and new ALBs using the `kubernetes.io/cluster/<cluster-id>` will be setup instead. This means a manual cleanup of old ALBs is required.

The cleanup is a one-time task and thus I did not put effort into adding it to the controller itself as it would be a significant change.

Fix #77 

Somewhat related to #95

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,8 +77,8 @@ On startup, the controller discovers the AWS resources required for the controll
 
 2. The Security Group
 
-    Lookup of the `kubernetes.io/cluster/<cluster-id>` tag of the Security Group matching the clusterID for the controller node and the
-    tag `aws:cloudformation:logical-id` matching the value `IngressLoadBalancerSecurityGroup`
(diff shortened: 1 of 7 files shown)


szuecs:
:+1:

mikkeloscar:
:+1:

PR #99, opened 2017-11-24
Add support for legacy KubernetesCluster tag

mikkeloscar:
This adds support for the legacy KubernetesCluster tag as a fallback
when detecting clusterID from the host instance.

This is to support kops which still use the old tag for the ec2
instances.

The new tag kubernetes.io/cluster/<cluster-id> will still be used and
expected by the ingress controller for the security group and the stack
that is created.

### DIFF
--- a/aws/ec2.go
+++ b/aws/ec2.go
@@ -10,14 +10,15 @@ import (
 )
 
 const (
-	defaultClusterID        = "unknown-cluster"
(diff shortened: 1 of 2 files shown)


szuecs:
:+1:

mikkeloscar:
:+1:

Issue #102, opened 2017-11-29
Support multiple subnets per AZ using Kubernetes AWS tags


szuecs:
As listed in https://github.com/kubernetes/kubernetes/blob/master/pkg/cloudprovider/providers/aws/aws.go#L85-L91

These tags should be added to Subnets by the cloud-admin or cluster-operator to select the Subnet in which to create the loadbalancer interfaces (kubernetes uses these for ELB creation for service type LoadBalancer):

    // TagNameSubnetInternalELB is the tag name used on a subnet to designate that
    // it should be used for internal ELBs
    const TagNameSubnetInternalELB = "kubernetes.io/role/internal-elb"

    // TagNameSubnetPublicELB is the tag name used on a subnet to designate that
    // it should be used for internet ELBs
    const TagNameSubnetPublicELB = "kubernetes.io/role/elb"

If we find public and private Subnets we should select them first by Tag and if nothing was found fall back to pick 3 sorted by name (?).

linki:
Possible solution: https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/103

szuecs:
Cool thanks @linki

cmcconnell1:
Hello, 'apologies if this is the wrong place, but hitting a snag here and I don't see any mention of subnet (or how to configure for specific subnets) in the main project readme.   For this cluster, I have two worker node pools each in a different AZ.

Just want to validate the solution is:  tag the subnets as specified above by @szuecs  ?

Details:
Seem to be stuck in a loop with the following error:
```A load balancer cannot be attached to multiple subnets in the same Availability Zone```

CloudFormation logs
```
16:11:53 UTC-0800	DELETE_COMPLETE	AWS::CloudFormation::Stack	opsdev-9df4348	
16:11:52 UTC-0800	DELETE_COMPLETE	AWS::ElasticLoadBalancingV2::TargetGroup	TG	
16:11:51 UTC-0800	DELETE_IN_PROGRESS	AWS::ElasticLoadBalancingV2::TargetGroup	TG	
16:11:51 UTC-0800	DELETE_COMPLETE	AWS::ElasticLoadBalancingV2::LoadBalancer	LB	
16:11:38 UTC-0800	DELETE_IN_PROGRESS	AWS::CloudFormation::Stack	opsdev-9df4348	The following resource(s) failed to create: [TG, LB]. . Delete requested by user.
[...]

PR #103, opened 2017-11-29
Limit to tagged ELB subnets if found

linki:
This limits the found subnets to the ones tagged with either `kubernetes.io/role/elb` or `kubernetes.io/role/internal-elb` in order to control which subnets are picked in case there are many.

If no subnet is tagged with either annotation then we assume there's no need to identify them and return the entire list. This is equivalent to the current behaviour.

Fixes: https://github.com/zalando-incubator/kube-ingress-aws-controller/issues/102

### DIFF
--- a/aws/adapter.go
+++ b/aws/adapter.go
@@ -340,6 +340,23 @@ func buildManifest(awsAdapter *Adapter) (*manifest, error) {
 	}
 
 	var (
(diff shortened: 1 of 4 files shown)


szuecs:
:+1:

linki:
👍

Issue #22, opened 2017-01-24
Support "internal" Application Load Balancers


hjacobs:
Support an annotation to create an `internal` ALB instead of `internet-facing`. The annotation for the Kubernetes ELB integration is `service.beta.kubernetes.io/aws-load-balancer-internal`. See https://github.com/kubernetes/kubernetes/blob/master/pkg/cloudprovider/providers/aws/aws.go#L78

hjacobs:
This is low priority for me (it's not needed in our Zalando context), but we should support it for completeness and other users.

mumoshu:
@hjacobs @jalonsoa Just get to see this issue. Can we close this as per #104 now?

mikkeloscar:
@mumoshu yes, this is solved as per #104 and #115. Thanks for pointing it out :)

PR #104, opened 2017-12-06
Add annotation to select scheme for elb creation

jalonsoa:
Related to the [Support "internal" Application Load Balancers](https://github.com/zalando-incubator/kube-ingress-aws-controller/issues/22) issue.

I add a zalando annotation to select internet-facing (default) or internal scheme to ELB

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,6 +125,7 @@ metadata:
   name: myingress
   annotations:
     zalando.org/aws-load-balancer-ssl-cert-domain: alt.name.org
+    zalando.org/aws-load-balancer-scheme: internal
 spec:
   rules:
   - host: test-app.example.org
@@ -160,6 +161,28 @@ spec:
           serviceName: test-app-service
           servicePort: main-port
 ```
+
+You can select the [Application Load Balancer Scheme ](http://docs.aws.amazon.com/elasticloadbalancing/latest/userguide/how-elastic-load-balancing-works.html#load-balancer-scheme)
+with an annotation like the one shown here:
+
+```yaml
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: myingress
+  annotations:
+    zalando.org/aws-load-balancer-scheme: internal
+spec:
+  rules:
+  - host: test-app.example.org
+    http:
+      paths:
+      - backend:
+          serviceName: test-app-service
+          servicePort: main-port
+```
+
+You can only select from internet-facing (default) and internal options.
 
 The new Application Load Balancers have a custom tag marking them as *managed* load balancers to differentiate them
 from other load balancers. The tag looks like this:
--- a/README.md
+++ b/README.md
@@ -125,7 +125,6 @@ metadata:
   name: myingress
   annotations:
     zalando.org/aws-load-balancer-ssl-cert-domain: alt.name.org
-    zalando.org/aws-load-balancer-scheme: internal
 spec:
   rules:
   - host: test-app.example.org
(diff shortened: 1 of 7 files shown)


kubernetes/adapter.go
```diff
@@ -92,6 +103,7 @@ func newIngressFromKube(kubeIngress *ingress) *Ingress {
 		namespace:      kubeIngress.Metadata.Namespace,
 		name:           kubeIngress.Metadata.Name,
 		hostName:       host,
+		scheme:         kubeIngress.getAnnotationsString(ingressSchemeAnnotation, "internet-facing"),
```
mikkeloscar: please use the AWS sdk constant here. We probably also should check if the specified value is valid and have a default if not.

README.md
```diff
@@ -125,6 +125,7 @@ metadata:
   name: myingress
   annotations:
     zalando.org/aws-load-balancer-ssl-cert-domain: alt.name.org
+    zalando.org/aws-load-balancer-scheme: internal
```
mikkeloscar: Please don't put it here, I think it's unrelated to this part of the docs.

README.md
```diff
@@ -161,6 +162,28 @@ spec:
           servicePort: main-port
 ```
 
+You can select the [Application Load Balancer Scheme ](http://docs.aws.amazon.com/elasticloadbalancing/latest/userguide/how-elastic-load-balancing-works.html#load-balancer-scheme)
```
mikkeloscar: nitpick: remove space before `]`

README.md
```diff
@@ -161,6 +162,28 @@ spec:
...
+  rules:
+  - host: test-app.example.org
+    http:
+      paths:
+      - backend:
+          serviceName: test-app-service
+          servicePort: main-port
+```
+
+You can only select from internet-facing (default) and internal options.
```
mikkeloscar: nitpick: please annotate the options with a code-block e.g. `internet-facing`

szuecs:
Thanks for your PR. 
It looks good in general, but one thing I don't like, but this could possibly be fixed with a small refactoring.
I would like to have no aws-sdk in the kubernetes package.

@jalonsoa would this be ok for you to do?

jalonsoa:
Hi @szuecs, 
  we add the import on this file to use the aws sdk constant (as @mikkeloscar recommended me, see his comments) for set default and validate values for the scheme annotation

  If you don't like this import .. where would you put the validation of the annotation ? Into the worker when we create the stack ? 

  So, we don't validate into the kubernetes adapter (don't use the import and don't use the "internet-facing" or "internal" strings), and validate into worker when we create or update the stack .... it's ok for you ?

szuecs:
@jalonsoa I just talked with @mikkeloscar . Thanks anyways for your PR :)

szuecs:
:+1:

mikkeloscar:
:+1:

jalonsoa:
Thanks !

so0k:
@mikkeloscar / @szuecs - something doesn't make sense with the annotation for elb scheme ...  (in the issue asking for this feature is not used by Zalando)... 

if we have 1 skipper DaemonSet and both internal/external LoadBalancers for our kube ingress, isn't is possible for attacker to hit the external LB with `Host: foo.internal.example.com` header set to access resources which are supposed to be accessible only on the internal LB?

We are currently patching this in our cluster by creating 2 kube-ingress-aws-controller deployments, where one will create load balancers hitting the hostPorts of our internal proxy and the other for our external proxy. In this case, it would more sense to set the default LB scheme as a command line flag when launching kube-ingress-aws-controller.

Your thoughts?

mikkeloscar:
@so0k that's a good point. Thinking about it now, I also don't really see when it's useful to have both internal and external by the same controller. As mentioned we still don't have a use-case for this in Zalando, so we never thought about how to actually use it :)

I think it would be perfectly fine to support this via a flag. You are welcome to open an issue/PR if you want to work on this :)

PR #111, opened 2017-12-17
Add sni support

### LINKED ISSUE #94: AWS Application Load Balancers Now Support Multiple TLS Certificates (SNI)
AWS now added SNI support for ALB. The ingress controller currently creates one ALB per certificate. We should update the ingress controller to create one ALB for multiple TLS certificates (if this is desired, maybe configurable?): https://aws.amazon.com/blogs/aws/new-application-load-balancer-sni/

mikkeloscar:
This adds SNI support by adding up to 25 certificates to a single ALB. If multiple ALBs exist with less than 25 certificates each then it will aggregate the certificates to as few ALBs as possible.

To support multiple certs this also fixes a bunch of other issues:

* **Changes how stacks are updated**. Before they were scheduled for update every hour (even if no updates were needed) and now it will check if there are any updates needed and update straight away. Without this change it would have taken 1 hour to add a new certificate to an existing ALB stack.
* **Changes cloudformation template** from a yaml template to use the library https://github.com/mweagle/go-cloudformation to generate the template using Go. This is needed because we want to dynamically add certificates to the stack, which is not possible with pure cloudformation.
* **Changes stack naming**. Before the name was something like `cluster-id-<hash-of-cert-arn>-scheme` but since we can now have 25 certs per stack the naming no longer makes sense. Instead it just gets the name `cluster-id-<uuid>`.

### Migration

This PR also takes care to ensure safe migration from stacks created by the `<=v0.5.0` version of the controller. The migration works like this:

1. Find old stacks with only one certificate.
2. Aggregate certs into one of the stacks (assuming less than 25 certs in total).
3. Mark old stacks to be deleted with a ttl on the existing cert the stack had defined.
4. Delete old stacks after the ttl has expired, allowing enough time for e.g. external-dns to move to the new stack.

Each certificate is defined on the stack with a tag `ingress:certificate-arn/<cert-arn>=ttl` where ttl is a timestamp for when the cert should be removed from the stack. This way it's possible to create multiple stacks in case there are more than 25 certs requested. And it's possible to later merge the stacks back to one when the demand for certs are less than 25 again.

Note: with this approach it's not possible to explicitly get an ALB that is unique per ingress/certificate we could add support for this either through a flag or as an annotation added to ingresses which want to have their own ALB, however this PR is already quite big so I would add these features separately.

Fix #94


aws/adapter.go
```diff
@@ -64,7 +64,7 @@ const (
 
 	nameTag = "Name"
 
-	certificateARNTag = "ingress:certificate-arn"
+	certificateARNsTag = "ingress:certificate-arns"
```
szuecs: this will break existing installations, please make it upgradeable and set a deprecated flag

aws/cf.go
```diff
@@ -20,12 +21,12 @@ const (
...
-	dnsName        string
-	scheme         string
-	targetGroupARN string
-	certificateARN string
-	tags           map[string]string
+	name            string
+	dnsName         string
+	scheme          string
+	targetGroupARN  string
+	CertificateARNs []string
```
szuecs: please do not change visibility here, better add a method, for example: Add(certarns ...string)

aws/cf.go
```diff
@@ -145,11 +142,75 @@ type healthCheck struct {
 	interval time.Duration
 }
 
+type Certificate struct {
+	CertificateArn string `yaml:"CertificateArn"`
+}
+
+// cfTemplate is an opauqe structure for unmarshaling a yaml cloudformation
```
szuecs: opauqe -> opaque

aws/cf.go
```diff
@@ -145,11 +142,75 @@ type healthCheck struct {
...
+		}
+		LB interface{} `yaml:"LB"`
+		TG interface{} `yaml:"TG"`
+	} `yaml:"Resources"`
+	Outputs interface{} `yaml:"Outputs"`
+}
+
+// injectCertificates injects a list of certificates into the cloudformation
+// template.
+func injectCertificates(cfTmpl string, certs []string) (string, error) {
```
szuecs: I am not happy with the solution parsing string, modify object and return string.
This looks more like a hack to me. We should have a type based object that can render to a cloudformation stack.

worker.go
```diff
@@ -138,10 +139,18 @@ func doWork(certsProvider certs.CertificatesProvider, awsAdapter *aws.Adapter, k
 	return nil
 }
 
-func buildManagedModel(certsProvider certs.CertificatesProvider, ingresses []*kubernetes.Ingress, stacks []*aws.Stack) map[string]*managedItem {
-	model := make(map[string]*managedItem)
+func buildManagedModel(certsProvider certs.CertificatesProvider, ingresses []*kubernetes.Ingress, stacks []*aws.Stack) []*managedItem {
```
szuecs: the changes in this function look more like we lack of abstraction and power of the abstraction.

szuecs:
:-1:

greenboxal:
Is this PR active? This is a super important feature for us as we use ingress for basically everything, and having one ALB per service is too expensive.

If not, I'm interested in finishing it.

szuecs:
@greenboxal I think we have not so much pressure as we provision wildcard certificates and mostly use the same ALB for all ingress already. Feel free to work on that.

mikkeloscar:
@greenboxal I have some half finished work locally, which I will try to push one of the next days.

mikkeloscar:
I have now updated this with all the changes I had locally. It now uses the github.com/mweagle/go-cloudformation library to generate a CF stack template instead of doing some weird yaml unmarshal/marshaling like it did before.

Note that this currently hasn't been tested and there are at least some issues I'm aware of:

* Naming of the stacks.
  * Currently we do this with a hash over the certificate, but since this can have 25 certs it no longer makes sense.
* When aggregating existing stacks into less stacks because of multiple certificates per stack/ALB then we need to do it in a way were the old stacks are kept for a bit during the migration. We have a default of 5 minutes right now (stackTTL).

Both of these things has to be done in a backwardcompatible way such that we can automatically migrate the single cert ALBs to multi cert ALBs without affecting applications behind those ingress ALBs.

@greenboxal If you want to go on from here, you are welcome to do so just let me know, otherwise I will work on this in CW04.

mikkeloscar:
@szuecs I have now updated the description explaining how it works.

I have tested the following scenarios:

* [x] Have two existing stacks from the previous version of the controller and switch to the new one.
* [x] Have a single stack with two certs and run a (custom) controller with a cert/stack limit of 1. This will split the stack in two.
* [x] Have two stacks with a cert each and run the controller with a cert/stack limit of 25 and see the stacks getting merged into one.
* [ ] **TODO** Test that all of this doesn't impact clients calling an endpoint behind ingress.

mikkeloscar:
@greenboxal This should now be done. You are very welcome to give it a try.

Here's an image built from this PR: `registry.opensource.zalan.do/teapot/kube-ingress-aws-controller-test:pr-111-15`

szuecs:
@mikkeloscar did you checked if we support  >2 ALB stacks?
With the current logic in buildManagedModel(), which is complex, I do not understand if this can work.
https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/111/files#diff-a693fca73f07436af23c207f04d5a5b7R194  
I think the whole logic should go into a method of `managedItem`.

worker.go
```diff
@@ -270,15 +320,17 @@ func updateIngress(kubeAdapter *kubernetes.Adapter, item *managedItem) {
...
-		if err := kubeAdapter.UpdateIngressLoadBalancer(ing, dnsName); err != nil {
-			if err != kubernetes.ErrUpdateNotNeeded {
-				log.Println(err)
+	for _, ingresses := range item.ingresses {
+		for _, ing := range ingresses {
+			if err := kubeAdapter.UpdateIngressLoadBalancer(ing, dnsName); err != nil {
+				if err != kubernetes.ErrUpdateNotNeeded {
+					log.Println(err)
+				} else {
+					log.Printf("updated ingress not needed %v with DNS name %q", ing, dnsName)
```
szuecs: please switch the error handling and change the message.

    if err == kubernetes.ErrUpdateNotNeeded {
        log.Printf("Ingress update not needed %v with DNS name %q", ing, dnsName)
    } else {
        log.Printf("Failed to update ingress: %v", err)
    }

szuecs:
I discussed offline with @mikkeloscar and checked the code with 4 eyes that it will support more than 2 stacks, too. (I dropped my old comment)

worker.go
mikkeloscar: Fixed!

szuecs:
:+1:

szuecs:
:+1:

mikkeloscar:
:+1:

PR #115, opened 2017-12-21
Group ingresses by ingress scheme

### LINKED ISSUE #114: Internal and Internet-Facing Aws-Load-Balancer-Scheme Cannot Coexist
I just experienced an issue where the Ingress Controller doesn't create an internal/internet-facing ALB stack when the other type already exists.

**How to Reproduce**:
1. Create an ingress with the internet-facing ingress with "zalando.org/aws-load-balancer-scheme: internal" annotation.
2. Create another ingress for the same domain (not the same host) but with "zalando.org/aws-load-balancer-scheme: internet-facing" annotation (or without any annotations at all, internet-facing being the default).
 
**Expected**: 
A new stack gets created as the first one cannot be used due to the fact that it's of a  different scheme.

**What Happens**:
The Ingress Controller doesn't create a new stack and reports the following in its logs: 

```
2017/12/21 20:55:53 Start polling sleep 30s
2017/12/21 20:56:23 Found 2 ingresses
2017/12/21 20:56:23 Found 1 stacks
2017/12/21 20:56:23 Have 1 models
2017/12/21 20:56:23 updated ingress not needed default/my-first-internal-app with DNS name "internal-k8s-the-rest-of-aws-alb-name"
2017/12/21 20:56:23 updated ingress not needed default/my-internet-facing-app with DNS name "internal-k8s-the-rest-of-aws-alb-name"
```
Note that the ALB DNS is that of the original ALB created for the first "internal" app.

I'm using the latest image for the controller:
registry.opensource.zalan.do/teapot/kube-ingress-aws-controller:latest

Here're the Ingress manifests:

 **Internal**:
```yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: "my-first-internal-app"
  annotations:
    zalando.org/aws-load-balancer-scheme: internal
  namespace: default
  labels:
    app: "my-first-internal-app"
spec:
  rules:
  - host: "my-first-internal-app.mydomain.com"
    http:
      paths:
      - backend:
          serviceName: "my-first-internal-app"
          servicePort: 80
```

**Internet-Facing**:
```yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: "my-internet-facing-app"
  annotations:
    zalando.org/aws-load-balancer-scheme: internet-facing
  namespace: default
  labels:
    application: "my-internet-facing-app"
spec:
  rules:
  - host: "my-internet-facing-app.mydomain.com"
    http:
      paths:
      - backend:
          serviceName: "my-internet-facing-app"
          servicePort: 80
```

I also tried omitting the annotation altogether for the seconds manifest.
Creating the internet facing ALB first leads to the same behaviour in reverse.

mikkeloscar:
This groups ingresses by loadbalancer scheme. Without grouping them the controller would create a single ALB for two ingresses with different schemes assuming the certificate was the same.

Fixes #114

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,13 +20,24 @@ This information is used to manage AWS resources for each ingress objects of the
 - Automatic discovery of SSL certificates
 - Automatic forwarding of requests to all Worker Nodes, even with auto scaling
 - Automatic cleanup of unnecessary managed resources
+- Support for internet-facing and internal load balancers
 - Can be used in clusters created by [Kops](https://github.com/kubernetes/kops), see our [deployment guide for Kops](deploy/kops.md)
 
 ## Upgrade
 
+### <v0.5.0 to >=v0.5.0
+
+Version `v0.5.0` introduced support for both `internet-facing` and `internal`
+load balancers. For this change we had to change the naming of the
+CloudFormation stacks created by the controller. To upgrade from v0.4.* to
+v0.5.0 no changes are needed, but since the naming change of the stacks
+migrating back down to a v0.4.* version will not be non-disruptive as it will
+be unable to manage the stacks with the new naming scheme. Deleting the stacks
+manually will allow for a working downgrade.
+
 ### <v0.4.0 to >=v0.4.0
 
-In versions before v.04.0 we used AWS Tags that were set by CloudFormation automatically to find
+In versions before v0.4.0 we used AWS Tags that were set by CloudFormation automatically to find
 some AWS resources.
 This behavior has been changed to use custom non cloudformation tags.
 
@@ -87,7 +98,7 @@ On startup, the controller discovers the AWS resources required for the controll
 
 2. The Security Group
 
-    Lookup of the `kubernetes.io/cluster/<cluster-id>` tag of the Security Group matching the clusterID for the controller node and `kubernetes:application` matching the value `kube-ingress-aws-controller` or as fallback for <v0.4.0
+    Lookup of the `kubernetes.io/cluster/<cluster-id>` tag of the Security Group matching the clusterID for the controller node and `kubernetes:application` matching the value `kube-ingress-aws-controller` or as fallback for `<v0.4.0`
     tag `aws:cloudformation:logical-id` matching the value `IngressLoadBalancerSecurityGroup` (only clusters created by CF).
 
 ### Creating Load Balancers
--- a/aws/adapter.go
+++ b/aws/adapter.go
@@ -229,9 +229,9 @@ func (a *Adapter) FindManagedStacks() ([]*Stack, error) {
 // from the Cluster ID and the certificate ARN (when available).
 // All the required resources (listeners and target group) are created in a transactional fashion.
 // Failure to create the stack causes it to be deleted automatically.
-func (a *Adapter) CreateStack(certificateARN string, scheme string) (string, error) {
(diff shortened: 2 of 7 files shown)


mikkeloscar:
@grifonas can you try this fix? It's available in this docker image:

```
registry.opensource.zalan.do/teapot/kube-ingress-aws-controller-test:pr-115-1
```

grifonas:
@mikkeloscar Trying...

grifonas:
@mikkeloscar Doesn't look like it's working.
Here's what I've got in the log:

```
2017/12/21 22:10:43 Start polling sleep 30s
2017/12/21 22:11:13 Found 2 ingresses
2017/12/21 22:11:13 Found 1 stacks
2017/12/21 22:11:13 Have 1 models
2017/12/21 22:11:13 creating "internet-facing" stack for certificate "arn:aws:acm:us-east-1:11111111111:certificate/bla-bla-bla" / ingress ["default/original-internet-facing-app"]
2017/12/21 22:11:13 updated ingress not needed default/original-internet-facing-app with DNS name "k8s-bla-bla-bla.us-east-1.elb.amazonaws.com"
2017/12/21 22:11:13 Start polling sleep 30s
```
As you can see, even though it's now showing the "creating message" it's incorrect.
Plus, it hasn't initiated any new CF stacks. =(

mikkeloscar:
@grifonas found a few more issues related to the load balancer schemes. Now it should all be fixed. You can try the image:

```
registry.opensource.zalan.do/teapot/kube-ingress-aws-controller-test:pr-115-3
```

grifonas:
Hi @mikkeloscar,
Looks like it did the job. Thanks!
Can this image be used in production safely or do you recommend waiting for the fix in the main image?

mikkeloscar:
> Can this image be used in production safely or do you recommend waiting for the fix in the main image?

@grifonas there's nothing special about the image, it's built the same way as the versioned images. So if you trust the changes in the PR it should be fine :)

szuecs:
After merge we have to increase the major version, because of the public interface changes.

@mikkeloscar do you see other possible breaking changes if we roll out this to running production clusters?

👍

mikkeloscar:
> After merge we have to increase the major version, because of the public interface changes.

Since this is not a library, but a standalone application I don't think we neccesarily need to bump the version because of the changes in the `aws` package. Unless someone is using it outside of this project?

@mikkeloscar do you see other possible breaking changes if we roll out this to running production clusters?

The changes are done in a way that it should not have an impact to upgrade. It will only create new stacks with the `-<scheme>` suffix if there are no existing stacks with the same scheme/certificate. However rolling back from this version to the previous one will not work as the old version wouldn't be able to understand the new stack name format.

I already successfully tested the upgrade a few times, but more testing wouldn't hurt ofc. :)

szuecs:
@mikkeloscar I would suggest to increase minor version if we can not easily rollback and have an additional change to the readme.md as we did before for the update to >= v0.4.0.
I bet users would also find it useful to have this behavior documented (referring to -<scheme>). I think we have no documentation regarding this, but I hope it's ok to request it before merge. :)

mikkeloscar:
@szuecs added some info about upgrade/downgrade to the README.

szuecs:
:+1:

szuecs:
@mikkeloscar you can merge :)

mikkeloscar:
:+1:

PR #117, opened 2018-01-01
Predictably select subnets

### LINKED ISSUE #108: stop delete create loop caused by 2 subnets
If an ALB can not be created, because 2 possible subnets could be attached to the ALB, the comment
https://github.com/zalando-incubator/kube-ingress-aws-controller/issues/102#issuecomment-350153375 shows that we get into delete create loop.

Desired behaviour: log error, delete and stop to create the ALB.

mikkeloscar:
This changes the logic of selecting subnets for the ALB to match the logic used
by the kube-controller-manager when selecting subnets for ELBs used for
services of type LoadBalancer.

This solves several problems:

1. The controller will no longer try to create an ALB with multiple subnets of
the same AZ. (Fixes #108)

2. If only some subnets are tagged with the elb role the controller will favour
those with the tag but not limit the selection to only those with a tag.

3. Favour using internal subnets for ALBs with the scheme `internal`.

Fix #108

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,18 @@ On startup, the controller discovers the AWS resources required for the controll
     Lookup of the `kubernetes.io/cluster/<cluster-id>` tag of the Security Group matching the clusterID for the controller node and `kubernetes:application` matching the value `kube-ingress-aws-controller` or as fallback for `<v0.4.0`
     tag `aws:cloudformation:logical-id` matching the value `IngressLoadBalancerSecurityGroup` (only clusters created by CF).
 
+3. The Subnets
+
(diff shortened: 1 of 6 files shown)


szuecs:
:+1:

szuecs:
lgtm for the code, but can you add docs for this, too?
Would be great, because our docs are not the best and we should make it better in every change we do.

I also stumbled about this test, I guess it is something to document:

 +			name: "should select first lexicographically subnet when two match a single zone",

mikkeloscar:
@szuecs Documented how subnets are discovered in the README.

szuecs:
:+1:

Raffo:
👍

Issue #116, opened 2017-12-24
Use Custom Tag for Discovery


grifonas:
Hi All,

The Ingress Controller is currently using the ```aws:autoscaling:groupName``` tag.

In our setup, however, some of the nodes are managed by a custom system rather than an autoscaling group.

Which leads me to the following questions:
1. Is it possible to customise the tag the Ingress Controller uses for discovery?

When trying to run the controller on a node not created by an AS group we're getting the following error:
```bash
2017/12/24 14:58:40 starting /bin/kube-ingress-aws-controller
2017/12/24 14:58:41 instance is missing the "aws:autoscaling:groupName" tag
```
2. Can the ingress-controller be ran as a daemon set? This will make sure that all the instances are discovered regardless of what autoscaling group (if any) they were created by.

szuecs:
Hi! Thanks for describing this issue. 

1. Not right now, but this is a doable enhancement we should tackle as sooner as better.
As you have written we use `aws:autoscaling:groupName` to identify the nodes for the targetGroup of the ALBs created by this ingress controller

2. This will not provide the value you assume I think. I think in general we should use a custom Tag to identify the target nodes, that we should be able to create via ASG or manual (automated by the user).

Does this makes sense to you?

mikkeloscar:
To add to what @szuecs already said so far we have assumed that all the instances attached to the ALB is in an autoscaling group because it makes it very simple for the controller. We can have AWS attach/dettach instances from the ALB automatically without worrying about this logic in the controller.

For your use case, and for the use case of using spot instances/fleets (#105) we will probably need to implement this logic ourselves as there currently is no automatic way like with ASGs.

So in my mind we would need to do something like this continuously:

1. Discover all instances belonging to the cluster (possibly filtering by some other tag to only select those that should be used by the ALB).
2. Attach all instances to the ALB
3. Remove instances from ALB which should no longer be attached e.g. because they are terminating. (Maybe AWS does something automatic here, I'm not completely sure how this works).
4.  Go back to 1 and start over.

grifonas:
Hi @szuecs  and @mikkeloscar!
Thanks very much for your responses.
What you both said makes perfect sense. 

I suppose one could simply replace the value for the "autoScalingGroupNameTag" variable in ec2.go with the value taken from an environment variable that could in turn be provided in the controller manifest, couldn't one? It can fall back to "aws:autoscaling:groupName" if nothing's provided.

[...]

PR #118, opened 2018-01-11
Multiple Auto Scaling Groups and Instances not in Auto Scaling Group management

dintel:
This change adds 2 new features (related to issue #116).

1. Support for multiple Auto Scaling Groups. If K8S cluster has multiple ASGs, kube-ingress-aws-controller will correctly handle them. To support this ASG of all nodes in K8S cluster is discovered to generate list of ASGs that should be updated. Kubernetes master nodes are skipped so they are not added to Target Groups.
2. Support for instances not in Auto Scaling Group (for example SpotInst). After fetching list of nodes from Kubernetes (without masters of course), instances which have no `aws:autoscaling:groupName` tag are saved in separate list. Instances from this list are added to Target Groups manually by a call to `RegisterTargets` ELBv2 API.


szuecs:
Thanks for your PR!

aws/adapter.go
```diff
@@ -399,3 +430,80 @@ func getTag(tags map[string]string, tagName string) (string, error) {
 	}
 	return "<missing tag>", ErrMissingTag
 }
+
+// Update list of ASGs that are updated with TargetGroup setting.
+func (a *Adapter) UpdateAutoScalingGroups(instancePrivateIps []string) error {
```
szuecs: Better name maybe: UpdateInstanceCaches

README.md
```diff
@@ -87,21 +89,25 @@ This is achieved using AWS CloudFormation. For more details check our [CloudForm
...
 It assumes that their lifecycle is external to the controller itself.
 
+Every poll cycle Kubernetes cluster is scanned for new nodes. Each new
+discovered node is scanned for Auto Scaling Group tag. Each Target
+Group created by this Ingress controller is then added to each Auto
+Scaling Group.
+If instance does not belong to Auto Scaling Group (does not have
+`aws:autoscaling:groupName` tag) it is stored in separate list of
+Single Instances. On each cycle instances on this list are registered
+as tartgets in all Target Groups managed by this controller.
```
szuecs: I think this part fits better into "Discovery", not?

dintel: I took it out of initial discovery phase and added to each cycle because otherwise newly added instances (after start of Ingress pod) would not be discovered and handled correctly.
I did everything I could to minimize number of calls to AWS APIs though.

aws/adapter.go
dintel: This function updates both instance caches and list of known ASGs.
Cache is used to reduce number of calls to DescribeInstances.
I assume that instances are not taken out of ASG (strictly speaking it's wrong, but IMHO rarely used in production).
May be call it UpdateAutoScalingGroupsWithInstanceCache ?

szuecs: I am fine with this

README.md
szuecs: ok

hjacobs:
@dintel thanks for the PR! The support for instances without ASG is exactly what we also need :smile:

szuecs:
@mikkeloscar I would prefer the EC2 way as you described. 

I did not have time to read the full PR, but I think the node finding should be in the aws package and not via kubernetes. 
In the end ALB can only target nodePorts, so I don't see a point of limiting us to kubernetes nodes. Less API calls the better. I would also like to have an option flag that is used as opt-in for ec2 nodes without ASG, if not done.

I try to do a full review next week.

dintel:
The reason for relying on Kubernetes API to fetch list of nodes was to stop relying on EC2 tags at all. By fetching nodes from Kubernetes we get a real list of nodes in Kubernetes cluster (without instances that are just starting or shutting down and are not ready yet). This list is not limited and includes nodes from all ASGs and even instances not in ASG.
The only thing that such approach would not allow is adding non-Kubernetes nodes to ALB as @szuecs wants and IMHO this can be important and very useful feature. So I will fix this ASAP. Should I open new PR when it's ready or just update this one?

mikkeloscar:
> So I will fix this ASAP. Should I open new PR when it's ready or just update this one?

It's not clear to me if you want to change the implementation to not use the Kubernetes API, or you will implement this while still using the Kubernetes API :)
Either way, pushing to this PR is fine I think, still related to this change.

szuecs:
@dintel if we detect a kubernetes node by kubernetes API, it does not have to be ready to serve traffic. The race condition is smaller, but in the end we all rely on active heartbeats from the application loadbalancer to detect healthy target in the target group.

Another idea that came into my mind, if you want to split the PR, then I would prefer in having one for the multi ASG and another one for the EC2 instance not being in an ASG.

Thanks for your work.

dintel:
@mikkeloscar I will change the implementation to use EC2 API instead of Kubernetes.

@szuecs Adding instances not in ASG is already part of the PR so let's continue with it.
Give me day or two and I will update this PR.

Just to clarify things a bit here is what I am going to do:

1. Remove part of code that retrieves nodes from Kubernetes (this will probably cancel all changes to kubernetes module).
2. During startup phase determine filter that is going to be used when calling EC2 DescribeInstances API:
  a. If `CUSTOM_TAGS_FILTER` environment variable is set it is expected to by in a form `Name=string,Values=string,string... Name=string,Values=string,string...`. This is very similar to what AWS CLI [describe-instances](https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-instances.html) --filter parameter accepts.
  b. If `CUSTOM_TAGS_FILTER` variable is not set or can not be parsed, the default is to fetch `KubernetesCluster` EC2 tag value for node where pod is running and setup filter for instances with same tag name and value and additionally tag `k8s.io/role/node` set.
3. During each cycle: 
  a. Load instances from EC2 using filter that was constructed during startup phase. Each instances information is cached so once instance information is loaded once, it is not reloaded.
  b. Every instance that belongs to ASG, it's ASG information is loaded and stored.
  c. Every instance that does not belong to ASG is stored in separate list.
  d. All known ASGs are updated to have Target Group associated with them.
  e. All known single instances are registered in Target Group.
  f. All instances that were removed from instance cache should be deregistered from target group.

Let me know what you think of this and do you have further suggestions.

dintel:
BTW, I am thinking whether we should manage instances in ASG and those that do not belong to any ASG separately? If single instances are registered and deregistered using ELBv2 API, then may be ASG part can be dropped and all instances should be treated as single instances. This change would simplify code and logic.
The only drawback I can think of in this case is that in case Ingress controller fails, Target Groups are not updated at all. Same is true for restarting. On the other side if something goes wrong, stopping Ingress controller would give an option to freeze state of Target Groups.
@mikkeloscar @szuecs what do you think?

szuecs:
@dintel, the problem with your suggestion about handling ASG instances the same as plain EC2 instances are the number of API calls you will have to do. We have to reduce them, because kubernetes is quite heavy in calls to AWS API and we get pretty often ratelimited and you do not want to manage the rollback on rate limits all over the place, that's why I wrote do the multi ASG first and do EC2 later. Please make sure you use CF to delete, update and create all resources in AWS. This is  a requirement for us running this in production for quite a long time now and improved the stability of the system a lot.

dintel:
I updated PR. In short here is what was done:

1. List of instances is now fetched from EC2 and not Kubernetes.
2. By default list of instances is filtered by `KubernetesCluster` and `k8s.io/role/node` tags, but this can be overridden by `CUSTOM_TAG_FILTER` environment variable.
3. Single Instances are now correctly registered and deregistered.
4. Ingress Controller pod can now run on Single Instance.
5. Documentation updated.

@szuecs @mikkeloscar should I remove nodes list support from kubernetes module as it is not used anymore?

szuecs:
@dintel yes please delete the code that is not in use.

mikkeloscar:
> By default list of instances is filtered by KubernetesCluster and k8s.io/role/node tags, but this can be overridden by CUSTOM_TAG_FILTER environment variable.

Just one note to this: The clusterID tag we should filter by should be `kubernetes.io/cluster/<cluster-id>` See more here: https://github.com/zalando-incubator/kubernetes-on-aws/pull/569

We should still also be able to filter on the `KubernetesCluster` tag but we consider this as legacy only to support kops setups.

dintel:
@mikkeloscar @szuecs all done.

README.md
```diff
@@ -87,21 +89,33 @@ This is achieved using AWS CloudFormation. For more details check our [CloudForm
...
+  filters are `tag:kubernetes.io/cluster/<cluster-id>=owned tag-key=k8s.io/role/node` where `<cluster-id>`
+  is determined from EC2 tags of instance on which Ingress Controller pod is started.
 
-1. The AutoScalingGroup
+Every poll cycle EC2 is queried with filters that were constructed during startup.
+Each new discovered instance is scanned for Auto Scaling Group tag. Each Target
+Group created by this Ingress controller is then added to each known Auto Scaling Group.
+Each Auto Scaling Group information is fetched only once when first node of it is discovered for first time.
+If instance does not belong to Auto Scaling Group (does not have `aws:autoscaling:groupName` tag) it is stored in separate list of
+Single Instances. On each cycle instances on this list are registered as tartgets in all Target Groups managed by this controller.
```
szuecs: tartgets -> targets

aws/adapter.go
```diff
@@ -119,6 +127,10 @@ func NewAdapter() (adapter *Adapter, err error) {
 		healthCheckInterval: DefaultHealthCheckInterval,
 		creationTimeout:     DefaultCreationTimeout,
 		stackTTL:            DefaultStackTTL,
+		autoScalingGroups:   make(map[string]*autoScalingGroupDetails),
+		instancesDetails:    make(map[string]*instanceDetails),
```
szuecs: instancesDetails is hard to read and spot the differences from instanceDetails type.
Maybe ec2Details?

aws/adapter.go
```diff
@@ -188,9 +200,58 @@ func (a *Adapter) InstanceID() string {
...
+	instances := make([]string, 0, len(a.singleInstances))
+	for id := range a.singleInstances {
+		instances = append(instances, id)
+	}
+	return instances
+}
+
+// RunningSingleInstances returns list of IDs of running instances that do
+// not belong to any Auto Scaling Group and should be managed manually.
+func (a *Adapter) RunningSingleInstances() []string {
```
szuecs: we could use the value receiver here

aws/adapter.go
```diff
@@ -188,9 +200,58 @@ func (a *Adapter) InstanceID() string {
...
+		if details.running {
+			instances = append(instances, id)
+		}
+	}
+	return instances
+}
+
+// ObsoleteSingleInstances returns list of IDs of instances that should be deregistered
+// from all Target Groups.
+func (a *Adapter) ObsoleteSingleInstances() []string {
```
szuecs: we could use the value receiver here

aws/adapter.go
```diff
@@ -188,9 +200,58 @@ func (a *Adapter) InstanceID() string {
...
+	return a.obsoleteInstances
+}
+
+// Get number of instances in cache.
+func (a *Adapter) CachedInstances() int {
```
szuecs: we could use the value receiver here

aws/adapter.go
```diff
@@ -188,9 +200,58 @@ func (a *Adapter) InstanceID() string {
...
+	return len(a.instancesDetails)
+}
+
+// Get EC2 filters that are used to filter instances that are loaded using DescribeInstances.
+func (a *Adapter) FiltersString() string {
```
szuecs: we could use the value receiver here.

aws/adapter.go
```diff
@@ -205,15 +266,38 @@ func (a *Adapter) FindManagedStacks() ([]*Stack, error) {
 	if err != nil {
 		return nil, err
 	}
+	return stacks, nil
+}
+
+// Update Auto Scaling Groups to have relevant Target Groups. Also
```
szuecs: exported functions should have godoc way of documentation, so you should start with "UpdateTargetGroupsAndAutoScalingGroups ..".

aws/adapter.go
```diff
@@ -399,3 +476,97 @@ func getTag(tags map[string]string, tagName string) (string, error) {
 	}
 	return "<missing tag>", ErrMissingTag
 }
+
+// Update list of known ASGs and EC2 instances.
```
szuecs: godoc: UpdateAutoScalingGroupsAndInstances updates the list of ...

aws/adapter.go
```diff
@@ -399,3 +476,97 @@ func getTag(tags map[string]string, tagName string) (string, error) {
...
+func (a *Adapter) UpdateAutoScalingGroupsAndInstances() error {
+	var err error
+	a.instancesDetails, err = getInstancesDetailsWithFilters(a.ec2, a.manifest.filters)
+	if err != nil {
+		return err
+	}
+
+	newSingleInstances := make(map[string]*instanceDetails)
+	for instanceId, details := range a.singleInstances {
```
szuecs: to be more golint compliant: instanceId -> instanceID

aws/adapter.go
dintel: Would not it have impact on performance? If we use value receiver here it will mean that each call to this method will make copy of Adapter struct, no?

aws/elbv2.go
```diff
@@ -0,0 +1,53 @@
...
+
+	for _, targetGroupARN := range targetGroupARNs {
+		input := &elbv2.DeregisterTargetsInput{
+			TargetGroupArn: aws.String(targetGroupARN),
+			Targets:        targets,
+		}
+
+		_, err := svc.DeregisterTargets(input)
+		if err != nil {
+			return fmt.Errorf("unable to deregister instances %q in target group %s: %v", instances, targetGroupARN, err)
```
szuecs: You will return error in case of AWS ratelimits the request, and the call to deregisterTargetsOnTargetGroups() from UpdateTargetGroupsAndAutoScalingGroups() will set the instance to "obsoleteInstances", but these are not deregistered correctly. I think you have to catch this here and introduce 2 error types, one for the case if you want to make these obsolete and the ratelimit one where you just want to go on and retry it in the next cycle.

aws/elbv2.go
```diff
@@ -0,0 +1,53 @@
...
+
+	for _, targetGroupARN := range targetGroupARNs {
+		input := &elbv2.RegisterTargetsInput{
+			TargetGroupArn: aws.String(targetGroupARN),
+			Targets:        targets,
+		}
+
+		_, err := svc.RegisterTargets(input)
+		if err != nil {
+			return fmt.Errorf("unable to register instances %q in target group %s: %v", instances, targetGroupARN, err)
```
szuecs: Reading https://docs.aws.amazon.com/elasticloadbalancing/latest/APIReference/API_RegisterTargets.html errors, it seems that we should not stop in the loop.
I would suggest log an error and do a continue. Any thoughts on this behaviour?

README.md
```diff
@@ -87,21 +89,33 @@ This is achieved using AWS CloudFormation. For more details check our [CloudForm
 The controller *will not* manage the security groups required to allow access from the Internet to the load balancers.
 It assumes that their lifecycle is external to the controller itself.
 
-### Discovery
+During startup phase EC2 filter is constructed as follows:
 
-On startup, the controller discovers the AWS resources required for the controller operations:
+* If `CUSTOM_TAG_FILTER` environment variable is set, it is expected to be in the form
+  of `FILTER_NAME=FILTER_VALUE1,FILTER_VALUE2... FILTER2_NAME=FILTER2_VALUE1,FILTER2_VALUE2... ...`.
```
szuecs: The triple dots does not makes sense to me (only the last to show you can add more filters).
I was reading parseFilters() comment and strings.Fields usage, which should be documented here in a user friendly way with an example how to use the custom filer ENV.
I suggest:

If `CUSTOM_TAG_FILTER` environment variable is set, it is expected to be in the form of `CUSTOM_TAG_FILTER="FILTER_NAME=FILTER_VALUE1,FILTER_VALUE2 FILTER2_NAME=FILTER2_VALUE1,FILTER2_VALUE2 ..."`.

szuecs:
@dintel thanks for the good code, there are on minor changes left.
- [ ] user documentation enhancement for the custom filter
- [ ] handling  ec2Service.DescribeInstances() case if you get more results than fits into the page. I think the best fix would be to use  DescribeInstancesPages() https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#EC2.DescribeInstancesPages

aws/elbv2.go
dintel: This error is logged and ignored in calling function (UpdateTargetGroupsAndAutoScalingGroups).

szuecs: But this would mean one ASG failure could block all the others, right?

dintel: No. Look at UpdateTargetGroupsAndAutoScalingGroups function. It will log error and continue to next ASG. Error won't stop loop over autoScalingGroups.

dintel:
Updated documentation and switched from DescribeInstances to DescribeInstancesPages.
Also renamed environment variable to make it simpler.
@szuecs I change pointer receiver to value receiver where you asked. IMHO other parts of code should be reviewed to change to value receiver where pointer receiver is not needed.

szuecs:
:+1:

aws/adapter.go
```diff
@@ -399,3 +476,102 @@ func getTag(tags map[string]string, tagName string) (string, error) {
...
+	}
+	return generateDefaultFilters(clusterId)
+}
+
+// Generate default EC2 filter for usage with ECs DescribeInstances call based on EC2 tags
+// of instance where Ingress Controller pod was started.
+func generateDefaultFilters(clusterId string) []*ec2.Filter {
+	return []*ec2.Filter{
+		{
+			Name:   aws.String("tag:" + clusterIDTagPrefix + clusterId),
```
mikkeloscar: It should not have a `tag:` prefix IMO

aws/ec2.go
```diff
@@ -19,8 +19,10 @@ const (
 	kubernetesCreatorValue     = "kube-ingress-aws-controller"
 	autoScalingGroupNameTag    = "aws:autoscaling:groupName"
 	runningState               = 16 // See https://github.com/aws/aws-sdk-go/blob/master/service/ec2/api.go, type InstanceState
+	stoppedState               = 80 // See https://github.com/aws/aws-sdk-go/blob/master/service/ec2/api.go, type InstanceState
```
mikkeloscar: I don't see this used anywhere?

aws/ec2mock_test.go
```diff
@@ -6,9 +6,12 @@ import (
 	"github.com/aws/aws-sdk-go/service/ec2/ec2iface"
 )
 
+const DIP_SPLIT_SIZE = 2
```
mikkeloscar: nit: please use camelCase for go variables e.g. `dipSplitSize` since it shouldn't be exported.

aws/ec2.go
dintel: This is used only in tests. But the constant is not testing specific, so I put it here.

aws/adapter.go
dintel: Not only should, but must have. You can check DescribeInstances documentation.

mikkeloscar: Ok I see, this I didn't know.

README.md
```diff
@@ -87,21 +89,42 @@ This is achieved using AWS CloudFormation. For more details check our [CloudForm
...
 It assumes that their lifecycle is external to the controller itself.
 
-### Discovery
+During startup phase EC2 filters are constructed as follows:
 
-On startup, the controller discovers the AWS resources required for the controller operations:
+* If `CUSTOM_FILTERS` environment variable is set, it is used to generate filters that are later used
+  to fetch instances from EC2.
+* If `CUSTOM_FILTERS` environment variable is not set or could not be parsed, then default
+  filters are `tag:kubernetes.io/cluster/<cluster-id>=owned tag-key=k8s.io/role/node` where `<cluster-id>`
```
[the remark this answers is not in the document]
dintel: We are running Kops 1.8 and I see that nodes has `k8s.io/role/node` tag, not `kubernetes.io/role/node`. Can it be that this tag was changed only recently?

dintel:
@szuecs documentation fixed and constant name are fixed.
About stack deletion and single instances - I just checked it on my cluster and single instances do not have to be detached. Stack deletion completes successfully without errors (both on Ingress Controller side and AWS side) and Target Group is removed correctly.
One last thing is default filter that uses `k8s.io/role/node` tag. IMO we should just check which of tags `k8s.io/role/node` or `kubernetes.io/role/node` is defined on instance that Ingress Controller is running and define filter to use correct tag. In case both tags are not defined, just skip this filter and filter only by `kubernetes.io/cluster/<cluster-name>`. Although this will add master nodes into ALB too, that should not be a problem and should be a sane default.

dintel:
@szuecs Tests failed because of expired certificate in kubernetes module testing data.

mikkeloscar:
> IMO we should just check which of tags k8s.io/role/node or kubernetes.io/role/node is defined on instance that Ingress Controller is running and define filter to use correct tag. In case both tags are not defined, just skip this filter and filter only by kubernetes.io/cluster/<cluster-name>. Although this will add master nodes into ALB too, that should not be a problem and should be a sane default.

I somewhat misunderstood how this works. I thought the tag would be something like `k8s.io/role/node=worker` which is the pattern we use. There seem to be no real standards but a few different "conventions" in kops and kubeadm. We try to follow what is done in kubeadm. Anyway, since it's a filter I'm fine with keeping it as is, then we can adjust the `CUSTOM_FILTER` for our setup and others can change it as they like.

dintel:
@mikkeloscar Tag name is `k8s.io/role/node` (or `kubernetes.io/role/node` as @szuecs suggested). Tag's value is not checked by default filter, only name. That's because value of that tag is '1', which has no useful meaning.
Full default filter looks like this `tag:kubernetes.io/cluster/<cluster-id>=owned tag-key=k8s.io/role/node`. This will filter nodes that has tag named `kubernetes.io/cluster/<cluster-id>` with value `owned` and tag named `k8s.io/role/node` regardless of it's value.

mikkeloscar:
@dintel 

Yup I do understand it now :) It's fine for me to keep it as is.

aws/elbv2.go
szuecs: Ok thanks to clarify this

szuecs:
@dintel cool thanks, I think we can this merge and fix the build with the TLS cert later in a different PR from ourselves.

szuecs:
:+1:

szuecs:
@dintel can you rebase on https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/119 this will fix the test

dintel:
@szuecs Rebased on master.

szuecs:
@dintel From my side is everything now done to be merged :)

szuecs:
:+1:

mikkeloscar:
:+1:

mikkeloscar:
@dintel thanks a lot for all the work.

We would like to take a bit of time and test it in our environment before we make a release. Until then an image can be found here for anyone who would like to try it out:

```
docker pull registry.opensource.zalan.do/teapot/kube-ingress-aws-controller:v0.5.0-2-g289884b
```

PR #123, opened 2018-02-07
[WIP] Update stacks when needed

mikkeloscar:
This PR aims to address two issues.

1. Stacks were scheduled to update by sending them to a channel and having a separate go routine triggering the updates every hour (by default). Instead of waiting one hour to do the update we can just check if an update is needed and schedule it right away. This is important for the SNI support (#111) as we want to add new certificates to existing ALBs and not wait an hour for it to propagate.

2. `FindManagedStacks` was only returning stacks in a completed state, this resulted in more stacks being created in cases where a stack update was triggered and some stacks were not returned from the `FindManagedStacks` call because they were in an update state. Instead it just returns all stacks and checks the state when deciding if they should be updated or not. I.e. stacks already being updated will not be updated again until they reach a complete state first.

I marked this **WIP** because I need to validate that this works correctly, but initial review is welcome :)

### DIFF
--- a/aws/cf.go
+++ b/aws/cf.go
@@ -21,6 +21,7 @@ const (
 // Stack is a simple wrapper around a CloudFormation Stack.
 type Stack struct {
 	name           string
+	status         string
 	dnsName        string
 	scheme         string
 	targetGroupARN string
@@ -46,6 +47,14 @@ func (s *Stack) CertificateARN() string {
 
 func (s *Stack) TargetGroupARN() string {
 	return s.targetGroupARN
+}
+
+// IsComplete returns true if the stack status is a complete state.
+func (s *Stack) IsComplete() bool {
+	if s == nil {
+		return false
+	}
+	return isComplete(s.status)
 }
 
 // IsDeleteInProgress returns true if the stack has already a tag
@@ -290,10 +299,8 @@ func getCFStackByName(svc cloudformationiface.CloudFormationAPI, stackName strin
 
 	var stack *cloudformation.Stack
 	for _, s := range resp.Stacks {
-		if isComplete(s.StackStatus) {
-			stack = s
-			break
-		}
+		stack = s
+		break
 	}
 	if stack == nil {
 		return nil, ErrLoadBalancerStackNotReady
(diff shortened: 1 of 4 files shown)


aws/cf.go
```diff
@@ -48,6 +49,14 @@ func (s *Stack) TargetGroupARN() string {
 	return s.targetGroupARN
 }
 
+// IsComplete returns true if the stack status is a complete state.
+func (s *Stack) IsComplete() bool {
+	if s == nil {
+		return false
+	}
+	return isComplete(s.status)
```
szuecs: I think it would be better to move the whole isComplete(string) body to this function, because it's only used if you have a Stack and this will reduce number of functions spread.

worker.go
```diff
@@ -175,8 +157,12 @@ func buildManagedModel(certsProvider certs.CertificatesProvider, ingresses []*ku
 		}
 		if item, ok := model[certificateARN+"/"+ingress.Scheme()]; ok {
 			item.ingresses = append(item.ingresses, ingress)
+			item.Update = true
 		} else {
-			model[certificateARN+"/"+ingress.Scheme()] = &managedItem{ingresses: []*kubernetes.Ingress{ingress}}
+			model[certificateARN+"/"+ingress.Scheme()] = &managedItem{
+				ingresses: []*kubernetes.Ingress{ingress},
+				Update:    true,
+			}
```
mikkeloscar: `Update` is always true here. This is wrong..

worker.go
```diff
@@ -175,8 +157,12 @@ func buildManagedModel(certsProvider certs.CertificatesProvider, ingresses []*ku
 		}
 		if item, ok := model[certificateARN+"/"+ingress.Scheme()]; ok {
 			item.ingresses = append(item.ingresses, ingress)
+			item.Update = true
```
szuecs: This Update = true is in both cases set. We have to lookup it somewhere to decide if there was a change that we now want to react on. Maybe we have a currentKnownState and check the new data to this one, but this would mean we are becoming stateful and are not stateless anymore. I don't see another way of doing it or trigger updates by time.Duration and additionally in the initial run.

mikkeloscar: The state is in the cloudformation stack as it has the certificate arn and scheme defined. We just need to check against that an decide if there was a change.

PR #125, opened 2018-02-18
Always tests against latest go stable version

mikkeloscar:
This is similar to what we do for the Dockerfile build. (we always use the latest stable version of go automatically).


mikkeloscar:
:+1:

linki:
👍

PR #127, opened 2018-02-21
Remove obsolete Target Groups from ASGs

### LINKED ISSUE #126: Cleanup old TG in ASGs
Sometimes we see ASGs failing to launch instances because it has Target Groups defined which no longer exists.

asg_fail_launch

We should make sure the controller always cleans up after itself.

mikkeloscar:
This ensures that obsolete Target Groups gets removed from the ASGs.

Fix #126


szuecs:
:+1:

mikkeloscar:
:+1:

PR #128, opened 2018-02-22
Make certificate TTL configurable. Set default to 30m

mikkeloscar:
This makes the certificate TTL (which indicate how long a certificate should be attached to an old load balancer) configurable.

It also changes the default time from 5 minutes to 30 minutes to ensure old ALBs are kept while DNS is migrated giving clients connecting to a service an ingress more time to migrate between ALBs.

### DIFF
--- a/controller.go
+++ b/controller.go
@@ -20,6 +20,7 @@ import (
 
 const (
 	defaultDisableSNISupport = false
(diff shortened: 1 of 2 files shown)


szuecs:
:+1:

mikkeloscar:
:+1:

PR #133, opened 2018-02-23
Make it possible to get a unique ALB for a single ingress resource

### LINKED ISSUE #129: Allow opt out of SNI ALB
#111 added support for aggregating ingresses with different Certificates under a single ALB using SNI.

Some clients might not support SNI and thus we should support opting out of an SNI ALB and create an a single ALB per certificate like before.

I suggest we do this with an annotation on the ingress object.

Something like:

```yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: myingress
  annotations:
    zalando.org/aws-load-balancer-sni: false
```

https://en.wikipedia.org/wiki/Server_Name_Indication

mikkeloscar:
This adds an annotation `zalando.org/aws-load-balancer-shared: "false"` which, when set will make sure to provision an ALB that is unique for a single ingress resources. This is useful if a single application in the cluster has clients which doesn't support SNI or if it's getting so much traffic that it should not share the ALB with other applications.

It works by having a `ingress:owner=<namespace>/<ingress-name>` tag on the stacks which are reserved for a single ingress. If the annotation is later removed from the ingress the stacks will be re-aggregated.

Fix #129

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,10 +241,11 @@ spec:
 
 You can only select from `internet-facing` (default) and `internal` options.
 
-By default the ingress-controller will aggregate all ingresses under a single
-Application Load Balancer (unless running with `-disable-sni-support`). If you
-like to provision an Application Load Balancer that is unique for an ingress
-you can use the annotation `zalando.org/aws-load-balancer-shared: "false"`.
+By default the ingress-controller will aggregate all ingresses under as few
+Application Load Balancers as possible (unless running with
+`-disable-sni-support`). If you like to provision an Application Load Balancer
+that is unique for an ingress you can use the annotation
+`zalando.org/aws-load-balancer-shared: "false"`.
 
 The new Application Load Balancers have a custom tag marking them as *managed* load balancers to differentiate them
 from other load balancers. The tag looks like this:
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -241,6 +241,11 @@ spec:
 
 You can only select from `internet-facing` (default) and `internal` options.
 
+By default the ingress-controller will aggregate all ingresses under a single
```
szuecs: single? if you have more than 25 you have 2 :)

mikkeloscar: changed it to "as few as possible" :)

kubernetes/adapter_test.go
```diff
@@ -18,6 +18,7 @@ func TestMappingRoundtrip(t *testing.T) {
 		hostName:       "bar",
 		scheme:         "internal",
 		certificateARN: "zbr",
+		shared:         true,
```
szuecs: It would be nice to have also a test case that shows that shared: false works

szuecs:
:+1:

mikkeloscar:
:+1:

PR #134, opened 2018-02-28
Add flag for enabling stack termination protection on ALB stacks

mikkeloscar:
Adds flag for enabling stack termination protection on ALB stacks.

https://aws.amazon.com/about-aws/whats-new/2017/09/aws-cloudformation-provides-stack-termination-protection/

This is to have a bit more protection against users terminating infrastructure stacks in the AWS console.

### DIFF
--- a/deploy/requirements.md
+++ b/deploy/requirements.md
@@ -224,6 +224,11 @@ Please also note that the worker nodes will need the right permission to describ
     "Effect": "Allow"
 },
 {
+    "Action": "cloudformation:Update*",
+    "Resource": "*",
+    "Effect": "Allow"
+},
+{
     "Action": "cloudformation:Delete*",
     "Resource": "*",
     "Effect": "Allow"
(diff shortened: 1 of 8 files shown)


deploy/requirements.md
```diff
@@ -223,6 +223,11 @@ Please also note that the worker nodes will need the right permission to describ
     "Resource": "*",
     "Effect": "Allow"
 },
+{
+    "Action": "cloudformation:Update*",
```
szuecs: Is this a new requirement or forgot?
If it's new please add also a section for updating to the new version in the README.md

mikkeloscar: It was missing for stack updates but also covers `cloudformation:UpdateTerminationProtection`. I think it's good enough that the requirements.md are updated so it will work whether you use this feature or not.

szuecs:
:+1:

mikkeloscar:
:+1:

PR #141, opened 2018-03-08
Don't update ASG when there are no targetgroups

### LINKED ISSUE #139: fix 400er while stack is in update and has no targetgroup
Logs showing these error messages, but it's not really an error. This will print these errors while creating the CF stack to create the ALB, but the targetgroup is not there yet.

    2018/03/08 14:26:16 UpdateTargetGroupsAndAutoScalingGroups() failed to attach target groups to ASG: ValidationError: 1 validation error detected: Value '[]' at 'targetGroupARNs' failed to satisfy constraint: Member must satisfy constraint: [Member must have length less than or equal to 511, Member must have length greater than or equal to 1, Member must satisfy regular expression pattern: [\u0020-\uD7FF\uE000-\uFFFD\uD800\uDC00-\uDBFF\uDFFF\r\n\t]*]
        status code: 400, request id: a964926a-22dc-11e8-9104-49c7bc84f33f

mikkeloscar:
When there are no TargetGroups we should not update the targets.

Fix #139

### DIFF
--- a/aws/adapter.go
+++ b/aws/adapter.go
@@ -283,6 +283,12 @@ func (a *Adapter) UpdateTargetGroupsAndAutoScalingGroups(stacks []*Stack) {
 	for i, stack := range stacks {
 		targetGroupARNs[i] = stack.targetGroupARN
 	}
+
+	// don't do anything if there are no target groups
+	if len(targetGroupARNs) == 0 {
+		return
+	}
+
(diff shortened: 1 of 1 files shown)


mikkeloscar:
:+1:

szuecs:
:+1:

PR #143, opened 2018-03-22
Add support for finding multiple certificates per ingress

### LINKED ISSUE #142: Find certificates for all hostnames
Since we added SNI support it would now be possible to find certificates matching all host names defined in an ingress resource and attach them to the ALB instead of being limited to a single certificate possibly only matching one of the host names.

To implement this we just need to change this to return a list of certificates and add all of them to the ALB stack: https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/master/worker.go#L241

With this users would not need to define the annotation: `zalando.org/aws-load-balancer-ssl-cert-domain`

mikkeloscar:
This adds support for finding all matching certificates for an ingress in case multiple hostnames are defined.

Right now it will fail to create an ALB for the ingress if the certificate lookup fails for just one of the hostnames specified in the ingress. I'm not sure if it's better to fail or do best effort by creating an ALB with just the certificates it found.

Thoughts on that?

Fix #142

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,8 +150,8 @@ On startup, the controller discovers the AWS resources required for the controll
 
 ### Creating Load Balancers
 
-When the controller learns about new ingress resources, it uses the host specified in it to automatically determine
-the most specific, valid certificate to use. The certificate has to be valid for at least 7 days. An example ingress:
+When the controller learns about new ingress resources, it uses the hosts specified in it to automatically determine
+the most specific, valid certificates to use. The certificates has to be valid for at least 7 days. An example ingress:
 
 ```yaml
 apiVersion: extensions/v1beta1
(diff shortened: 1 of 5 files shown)


szuecs:
@mikkeloscar Maybe it makes sense to have a default behaviour which you can change with a flag.

hjacobs:
@mikkeloscar what was the behavior before (where only one cert per ingress was supported)? If it was happily creating the ALB as long as one cert was found then we should keep it like this (i.e. not fail).

szuecs:
@mikkeloscar Regarding your question to fail or not to fail:
In general it makes sense not to fail, in my opinion. The only problem is if for whatever reason we get no valid certs and destroy a current working ALB. Maybe you can catch this kind of "disaster" error.

mikkeloscar:
Thanks, I will change it to only fail if no certificates can be found and otherwise do best effort. Which is kinda how it works now.

hjacobs:
@mikkeloscar SGTM

mikkeloscar:
Now it will not fail if it finds at least one certificate.

Another thing. Should we remove support for the `zalando.org/aws-load-balancer-ssl-cert-domain` annotation? Right now it has the functionality that you can use it to limit the certificate lookup to a single domain, but I'm not sure this is very useful?

If we remove support for it, existing ingresses which specifies this annotation will potentially get an extra certificate attached to the ALB, but I don't think this is a problem.

What do you think?

szuecs:
@mikkeloscar I agree, that `zalando.org/aws-load-balancer-ssl-cert-domain` might be not a feature anymore, but I doubt this can be discussed in a separate issue. We can ask the community and give them some time to tell us if they find this useful or not.

szuecs:
👍

Issue #38, opened 2017-03-28
Use kubernetes.io/ingress.class annotation to be able to use with other controllers


ese:
If the ingress resource has the annotation only take it in account if its match with the controller class. For example `kubernetes.io/ingress.class: alb`

hjacobs:
I would definitely make this configurable (i.e. only require the annotation if it was enabled via command line flag): as a user, I don't want to add an unnecessary annotation if I'm only using one Ingress controller in the whole cluster.

mikkeloscar:
@hjacobs as I understand the issue is to only respect the annotation if it's there, if it's not there the ingress controller just takes full ownership.

Still this could be enabled/disabled with a flag.

hjacobs:
@mikkeloscar right, after reading @ese's text again I understand it in the same way :smile:

linki:
AppsCode's [Voyager](https://github.com/appscode/voyager) has a good approach. Like Mikkel describes and also configurable.

https://github.com/appscode/voyager#running-voyager-alongside-with-other-ingress-controller

lmineiro:
What value should we adopt for this controller? `alb` like suggested?

hjacobs:
We have one potential problem with the annotation: both the `kube-ingress-aws-controller` and Skipper need to listen on changes (in our Zalando setup). Skipper might just read ALL Ingress rules (ignoring the annotation), but otherwise we need to find a way of having multiple values for the annotation (to enable/disable `kube-ingress-aws-controller` and `skipper` separately).

szuecs:
@hjacobs Is it a problem if skipper serves also traffic configured by other ingress-controllers?
I think not, so skipper could just create all rules and if it does not get traffic for one rule, because someone use an ingress-controller that routes traffic to f.e. nginx, there is no problem.

lmineiro:
What @szuecs just mentioned is an important point that I've also been insisting.
The controller deliver things that are pretty standard - A LB with all the worker nodes as members and a DNS name. This can be used as a contract for what it delivers.
I'd even encourage other competing solutions to pick up that contract and transport traffic to applications inside a Kubernetes cluster.
We rely on Skipper and I'm confident this is the right thing for us but an explicit combination of the controller and Skipper should not be required.

jhohertz:
[...]

PR #158, opened 2018-05-05
Initial work for zalando-incubator/kube-ingress-aws-controller#38

jhohertz:
Implements a new flag -ingress-class-filter which if specified is
interpreted as a comma seperated list of values which will be used
as a filter to match kubernetes.io/ingress.class annotations. If
the annotation is not set, or does match, the ingress will be hidden
from the ingress controller and not acted upon.

As of this commit the work passes a 'make test' but requires
further testing.

### DIFF
--- a/kubernetes/adapter.go
+++ b/kubernetes/adapter.go
@@ -9,6 +9,7 @@ import (
 
 type Adapter struct {
 	kubeClient client
+	ingressFilters []string
 }
 
 var (
@@ -165,7 +166,7 @@ func newIngressForKube(i *Ingress) *ingress {
 }
 
 // NewAdapter creates an Adapter for Kubernetes using a given configuration.
-func NewAdapter(config *Config) (*Adapter, error) {
+func NewAdapter(config *Config, ingressClassFilters []string) (*Adapter, error) {
 	if config == nil || config.BaseURL == "" {
 		return nil, ErrInvalidConfiguration
 	}
@@ -173,7 +174,7 @@ func NewAdapter(config *Config) (*Adapter, error) {
 	if err != nil {
 		return nil, err
 	}
-	return &Adapter{kubeClient: c}, nil
+	return &Adapter{kubeClient: c, ingressFilters: ingressClassFilters}, nil
 }
 
 // ListIngress can be used to obtain the list of ingress resources for all namespaces.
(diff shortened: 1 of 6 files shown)


kubernetes/adapter.go
```diff
@@ -182,10 +183,26 @@ func (a *Adapter) ListIngress() ([]*Ingress, error) {
...
+				}
+			}
+		}
+	} else {
+		ret = make([]*Ingress, len(il.Items))
+		for i, ingress := range il.Items {
+			ret[i] = newIngressFromKube(ingress)
+		}
 	}
+	fmt.Println(ret)
```
szuecs: cleanup :)

szuecs:
@jhohertz besides this cleanup I am happy, thanks!

jhohertz:
Please hold any merging, I am running into some issues in testing I need to resolve. (It's not seeing any ingresses, regardless of parameters)

jhohertz:
I've pushed up two small changes, one just logs the ingress class filters at startup as other values get logged.

The other is the bug I saw, where if we didn't set the new param, it wasn't behaving as before, but trying to take a new path (Apparently go's string split on an empty string does not yield a zero sized array, which messed up the test I was doing.... now I check at init for a zero sized filter string, and pass a zero sized array explicitly)

From my perspective this is now ready to merge.

jhohertz:
Test build of container available at: viafoura/kube-ingress-aws-controller:v0.6.8-5-g6321660

kubernetes/adapter.go
```diff
@@ -165,15 +167,21 @@ func newIngressForKube(i *Ingress) *ingress {
...
 	if err != nil {
 		return nil, err
 	}
-	return &Adapter{kubeClient: c}, nil
+	return &Adapter{kubeClient: c, ingressFilters: ingressClassFilters}, nil
+}
+
+// Get ingress class filters that are used to filter ingresses acted upon.
+func (a *Adapter) IngressFiltersString() string {
+        result := fmt.Sprintf("%s", strings.Join(a.ingressFilters, ","))
```
szuecs: you don't need Sprintf here, just pass strings.Join() into strings.TrimSpace() :)

szuecs:
@jhohertz only a tiny change left. :)
Super awesome, thanks!

kubernetes/adapter.go
jhohertz: Will adjust, thanks.

szuecs:
:+1:

hjacobs:
:+1:

PR #162, opened 2018-05-15
Add a workaround for CloudFormation update issues

aermakov-zalando:
A bug in how CloudFormation processes updates for LBs can result in complete misconfiguration. The `AWS::ElasticLoadBalancingV2::Listener` resource must have exactly one certificate in the CertificateArn property, which is then treated as the default certificate. All the rest are managed in a separate resource of type `AWS::ElasticLoadBalancingV2::ListenerCertificate`. Unfortunately, the update is not performed atomically. CloudFormation updates the `Listener` first, and only after it's done it creates a new `ListenerCertificate` resource and deletes the old one.

However, if the previously default certificate swaps positions with one from the `ListenerCertificate` list, the resulting race conditions in AWS will result in a completely broken configuration. One of the certificates will be completely missing from the LB and you might or might not get update errors on the CF stack.

Implement the workaround suggested by AWS support, which is to change the name of the `ListenerCertificate` resource instead of updating an existing one. To make the resource name predictable, certificate ARNs are now sorted and their hash is used in the resource name. A nice side effect is that the default certificate will always be a wildcard one instead of an arbitrary one.

Additionally, the first certificate is now present in both `Listener` and `ListenerCertificate` resources. This doesn't seem to break anything, but will stop CloudFormation from deleting the default certificate if it stops being the default one.

### DIFF
--- a/aws/cf_template.go
+++ b/aws/cf_template.go
@@ -6,7 +6,20 @@ import (
 	"time"
 
 	"github.com/mweagle/go-cloudformation"
+	"sort"
+	"crypto/sha256"
 )
+
+func hashARNs(certARNs []string) []byte {
+	hash := sha256.New()
+
+	for _, arn := range certARNs {
+		hash.Write([]byte(arn))
+		hash.Write([]byte{'\000'})
+	}
+
+	return hash.Sum(nil)
+}
 
 func generateTemplate(certs map[string]time.Time, idleConnectionTimeoutSeconds uint) (string, error) {
 	template := cloudformation.NewTemplate()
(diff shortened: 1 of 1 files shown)


szuecs:
What if you have a second wildcard certifcate created that is lexicographically sorted before the first one?
For example you have `*.foo.example.org` and later you add `*.bar.example.org`. In this situation it will break again because of the order. I am fine with this but we should document it.

mikkeloscar:
@aermakov-zalando @szuecs and this should not really break as I understand it right? Because the "main" certificate will be stored in the `ListenerCertificate` list so it will be available even if the default is replaced by another wildcard cert.

aermakov-zalando:
@mikkeloscar @szuecs If we want to implement some logic to determine the best default certificate we can do it later, but this change shouldn't break anything.

PR #169, opened 2018-06-04
filter subnets on cluster ID tag before consideration

### LINKED ISSUE #167: Tag filtering for subnets is apparently not working right?
Sorry submitted this before I put a description in...

We are seeing the controller select subnets seemingly at random... where tags do not match the filter, causing cross-contamination between clusters, and in some cases selecting subnets not even tagged for k8s at all.

Just starting to gather details on this now.

jhohertz:
This fixes #167 by adding a filter when collecting subnet IDs for the manifest structure.

### DIFF
--- a/aws/ec2.go
+++ b/aws/ec2.go
@@ -151,7 +151,7 @@ func findFirstRunningInstance(resp *ec2.DescribeInstancesOutput) (*ec2.Instance,
 	return nil, ErrNoRunningInstances
 }
 
-func getSubnets(svc ec2iface.EC2API, vpcID, clusterId string) ([]*subnetDetails, error) {
+func getSubnets(svc ec2iface.EC2API, vpcID, clusterID string) ([]*subnetDetails, error) {
 	params := &ec2.DescribeSubnetsInput{
 		Filters: []*ec2.Filter{
 			{
@@ -181,7 +181,7 @@ func getSubnets(svc ec2iface.EC2API, vpcID, clusterId string) ([]*subnetDetails,
 			return nil, err
 		}
 		tags := convertEc2Tags(sn.Tags)
-		if _, ok := tags[clusterIDTagPrefix + clusterId]; ok {
+		if _, ok := tags[clusterIDTagPrefix + clusterID]; ok {
 			ret = append(ret, &subnetDetails{
 				id:               subnetID,
 				availabilityZone: az,
--- a/aws/ec2.go
+++ b/aws/ec2.go
@@ -181,8 +181,7 @@ func getSubnets(svc ec2iface.EC2API, vpcID, clusterId string) ([]*subnetDetails,
 			return nil, err
 		}
 		tags := convertEc2Tags(sn.Tags)
-		_, subnetHasClusterTag := tags[clusterIDTagPrefix + clusterId]
-		if subnetHasClusterTag {
+		if _, ok := tags[clusterIDTagPrefix + clusterId]; ok {
 			ret = append(ret, &subnetDetails{
 				id:               subnetID,
 				availabilityZone: az,
--- a/aws/ec2.go
+++ b/aws/ec2.go
@@ -151,7 +151,7 @@ func findFirstRunningInstance(resp *ec2.DescribeInstancesOutput) (*ec2.Instance,
 	return nil, ErrNoRunningInstances
 }
 
-func getSubnets(svc ec2iface.EC2API, vpcID string) ([]*subnetDetails, error) {
+func getSubnets(svc ec2iface.EC2API, vpcID string, clusterId string) ([]*subnetDetails, error) {
 	params := &ec2.DescribeSubnetsInput{
 		Filters: []*ec2.Filter{
 			{
@@ -160,6 +160,10 @@ func getSubnets(svc ec2iface.EC2API, vpcID string) ([]*subnetDetails, error) {
 					aws.String(vpcID),
 				},
 			},
+                        {
+                                Name:   aws.String("tag:" + clusterIDTagPrefix + clusterId),
+                                Values: []*string{aws.String(resourceLifecycleOwned)},
+                        },
 		},
 	}
 	resp, err := svc.DescribeSubnets(params)
--- a/aws/ec2.go
+++ b/aws/ec2.go
@@ -160,10 +160,6 @@ func getSubnets(svc ec2iface.EC2API, vpcID string, clusterId string) ([]*subnetD
 					aws.String(vpcID),
 				},
 			},
-                        {
-                                Name:   aws.String("tag:" + clusterIDTagPrefix + clusterId),
-                                Values: []*string{aws.String(resourceLifecycleOwned)},
-                        },
 		},
 	}
 	resp, err := svc.DescribeSubnets(params)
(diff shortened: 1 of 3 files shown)


szuecs:
:+1:

aws/ec2.go
```diff
@@ -160,6 +160,10 @@ func getSubnets(svc ec2iface.EC2API, vpcID string) ([]*subnetDetails, error) {
 					aws.String(vpcID),
 				},
 			},
+                        {
+                                Name:   aws.String("tag:" + clusterIDTagPrefix + clusterId),
+                                Values: []*string{aws.String(resourceLifecycleOwned)},
```
mikkeloscar: This is a bit too strict for us as we share subnets with the tag `kubernetes.io/cluster/<id>=shared`.

We could add `shared` as another possible value, but to be backwards compatible we should also support the current case were no cluster id is defined on the subnets.

mikkeloscar:
:-1:

aws/ec2.go
jhohertz: Perhaps it's better to just look for the tag key vs. using the filter. That's what the code this logic was based on is doing, this reference is just above the block referenced in the comments of the code:

https://github.com/kubernetes/kubernetes/blob/65efeee64f772e0f38037e91a677138a335a7570/pkg/cloudprovider/providers/aws/aws.go#L2923-L2928

I will have a look at using that approach instead.

mikkeloscar: @jhohertz yup, that makes sense, also because the logic then is the same for Kubernetes services of type load balancer, which is why we modeled the logic after that (but obviously didn't go the full way as you found).

Thanks for the effort!

jhohertz:
@mikkeloscar I have a new version just pushed to this branch which changes how the filtering works a bit, will only look to the tag key name, and work with either 'shared' or 'owned' (or any other) value.

aws/ec2.go
```diff
@@ -151,7 +151,7 @@ func findFirstRunningInstance(resp *ec2.DescribeInstancesOutput) (*ec2.Instance,
 	return nil, ErrNoRunningInstances
 }
 
-func getSubnets(svc ec2iface.EC2API, vpcID string) ([]*subnetDetails, error) {
+func getSubnets(svc ec2iface.EC2API, vpcID string, clusterId string) ([]*subnetDetails, error) {
```
szuecs: nitpick:
`func getSubnets(svc ec2iface.EC2API, vpcID string, clusterId string) ([]*subnetDetails, error) {`
-> `func getSubnets(svc ec2iface.EC2API, vpcID, clusterId string) ([]*subnetDetails, error) {`

aws/ec2.go
```diff
@@ -172,20 +172,23 @@ func getSubnets(svc ec2iface.EC2API, vpcID string) ([]*subnetDetails, error) {
...
 			return nil, err
 		}
 		tags := convertEc2Tags(sn.Tags)
-		ret[i] = &subnetDetails{
-			id:               subnetID,
-			availabilityZone: az,
-			public:           isPublic,
-			tags:             tags,
+		_, subnetHasClusterTag := tags[clusterIDTagPrefix + clusterId]
+		if subnetHasClusterTag {
```
szuecs: use idiomatic Go:

`if _, ok := tags[clusterIDTagPrefix + clusterId]; ok { ... }`

aws/ec2.go
mikkeloscar: Also nit: use `clusterID` instead of `clusterId` :)

hjacobs:
@jhohertz thanks for your contribution! This is highly appreciated :smile:

aws/ec2.go
```diff
@@ -151,7 +151,7 @@ func findFirstRunningInstance(resp *ec2.DescribeInstancesOutput) (*ec2.Instance,
 	return nil, ErrNoRunningInstances
 }
 
-func getSubnets(svc ec2iface.EC2API, vpcID string) ([]*subnetDetails, error) {
+func getSubnets(svc ec2iface.EC2API, vpcID, clusterId string) ([]*subnetDetails, error) {
```
mikkeloscar: nit: could we call the variable `clusterID` to be consistent :)

hjacobs:
@jhohertz as you get all the "nitpick" comments, it's a good opportunity for me to introduce Stackenblochen: https://www.youtube.com/watch?v=d6wd78jMHDk :smirk:

szuecs:
:+1:

mikkeloscar:
This lgtm, but this will change the behavior to require tagged subnets, which is not yet required in Kubernetes services: https://github.com/kubernetes/kubernetes/blob/v1.10.3/pkg/cloudprovider/providers/aws/aws.go#L3009

We should either support non-tagged subnets as a fallback like before (I would prefer this). Or at least mention this change in the README. WDYT @szuecs ?

szuecs:
@mikkeloscar most preferable seems to be adding a fallback and add a warning + comment with the link https://github.com/kubernetes/kubernetes/blob/v1.10.3/pkg/cloudprovider/providers/aws/aws.go#L3009.

jhohertz:
Just want to check before trying to implement this on approach, a few possibilities:

1.  In the loop I build two possible return lists, one filtered, the other without. If the filtered list come up with length > 0, that gets returned, else we provide the unfiltered list (and log warnings, but would need to import log where it not there today)

2. Variant on the above, I put a flag as a param to getSubnets to do the filtering or not, and the adapter that calls it will be responsible to first call with the flag set, and then fallback to calling it again without it set. The adapter is already logging things, so no new imports to log the warning.

3. I add a `--strict-subnets` flag, and add an or to the conditional wrapping the append that will let it always append if not set, letting the existing behaviour continue to be default, and the new must be explicitly asked for.

I kind of like the second approach if you want the fallback, or the third if you'd prefer a more explicit ask of the user for the new behaviour.

Thanks for all the feedback!

mikkeloscar:
Option 1 sounds best to me.

jhohertz:
Here is option 1. I can't easily test the fallback, but have tried to update the unit tests to cover both scenarios.

jhohertz:
One sec, just noticed some space vs tabs formatting issue, will force a push w/ that fixed.

szuecs:
:+1:

aws/ec2.go
```diff
@@ -181,15 +183,28 @@ func getSubnets(svc ec2iface.EC2API, vpcID string) ([]*subnetDetails, error) {
...
+				tags:             tags,
+			})
+		}
 	}
-	return ret, nil
-
+	// Fall back to full list of subnets if none matching expected tagging are found, with a stern warning
+	// https://github.com/kubernetes/kubernetes/blob/v1.10.3/pkg/cloudprovider/providers/aws/aws.go#L3009
+	if len(retFiltered) == 0 {
+		log.Printf("No tagged subnets found; considering all subnets. This is likely to be an error in a future versions.")
```
hjacobs: nitpick: "error in a future versions" --> "error in future versions"

hjacobs:
LGTM, see my nitpick with regards to the log message

hjacobs:
:+1:

hjacobs:
@jhohertz thanks again!

jhohertz:
You too @hjacobs , @mikkeloscar and @szuecs !

mikkeloscar:
:+1:

Issue #170, opened 2018-06-05
Allow annotation to define default cert


so0k:
Currently, the default cert is decided by picking the first certificate after sorting all matching certificates.

For my set up, I have wildcard certificates of several domains and ingress manifests for all of those domains.

I recently added a naked domain ingress request, and for this particular DNS record I changed the wildcard cert to have a SAN for the naked domain. It seems that there are validation errors if this certificate is not the default of the ALB (I don't quite understand yet why that's the case).

Would this annotation make sense?

so0k:
I guess I could use `zalando.org/aws-load-balancer-shared: false` - but currently only one ingress uses that naked domain.. the other domains are using subdomains and seem to work fine

hjacobs:
I'm not completely sure if I understand the scenario. What do you mean with "default" certificate for the ALB? I tried to find such property in the [ALB Listener Cloud Formation reference](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticloadbalancingv2-listener.html), but could not find it.

mikkeloscar:
> I recently added a naked domain ingress request, and for this particular DNS record I changed the wildcard cert to have a SAN for the naked domain. It seems that there are validation errors if this certificate is not the default of the ALB (I don't quite understand yet why that's the case).

Let's try to understand this first before we add workarounds. Can you give some more information about what your certs look like and what type of validation error you are getting?

We have found bugs with ALB/Certs/Cloudformation which have been confirmed by AWS, so if there is something like this we should understand it and reach out to them.


@hjacobs default cert is the one specified here: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticloadbalancingv2-listener.html#cfn-elasticloadbalancingv2-listener-certificates all the rest is under: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticloadbalancingv2-listenercertificate.html

so0k:
All my Certs are wildcard issued through ACM:

```
*.foo.com
*.bar.com
```

Everything was working fine with ingress requests like:

```yaml
ingress:
  hosts:
  - a.foo.com
ingress:
  hosts:
  - b.foo.com
ingress:
  hosts:
  - a.bar.com
```

Yesterday, I added an ingress request for `www` and naked domain for one of the TLDs

```
ingress:
  hosts:
  - www.qux.com
[...]

PR #171, opened 2018-06-05
remove zalando.org/aws-load-balancer-ssl-cert-domain

### LINKED ISSUE #149: Removing zalando.org/aws-load-balancer-ssl-cert-domain annotation
@mikkeloscar and I agree on that the annotation `zalando.org/aws-load-balancer-ssl-cert-domain` might not be a feature anymore.
We want to ask the community if you agree or disagree and if you disagree explain why. See [this comment](https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/143#issuecomment-380913956) for why it's not needed anymore.

szuecs:
remove zalando.org/aws-load-balancer-ssl-cert-domain annotation

This will fix #149

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,36 +171,7 @@ spec:
 The Application Load Balancer created by the controller will have both an HTTP listener and an HTTPS listener. The
 latter will use the automatically selected certificates.
 
-Alternatively, you can specify a domain used for automatically selecting a
-certificate with an annotation like the one shown below. This is useful if you
-have multiple host rules defined and want to only select one certificate for a
-specific hostname. Generally it's better to let the controller select all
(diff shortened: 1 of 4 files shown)


szuecs:
:+1:

szuecs:
:+1:

hjacobs:
What about #170, would the annotation help in that case? (Just making sure we don't delete stuff which is still valuable)

szuecs:
@hjacobs the name would need to be changed and reuse it to do something else is also error prone.
Additionally #170 needs to change the cf stack and we need to think about what to do in case of a changed default which does not work as we tested it.

PR #177, opened 2018-06-20
Only manage TGs that are owned by the controller

### LINKED ISSUE #165: Controller deletes other target groups from autoscaling group
Controller delete target groups from autoscaling group which are not managed by itself. This avoid have other stuff working with target groups managed by other components.

How to reproduce:
  * Create a target group
  * Add target group to the nodes autoscaling group

What happens:
  * kiac delete the target group from the autoscaling groups

What expect:
  * kiac maintain and don't delete stuff which is not managed by itself
  * kiac delete garbage target group looking for a label to know that was self-managed

mikkeloscar:
This makes sure that the controller only detaches target groups from ASGs when these target groups are managed/owned by the controller. The ownership is determined based on the clusterID and controllerID tags on the target groups.

Fix #165

### DIFF
--- a/aws/adapter.go
+++ b/aws/adapter.go
@@ -314,9 +314,14 @@ func (a *Adapter) UpdateTargetGroupsAndAutoScalingGroups(stacks []*Stack) {
 		return
 	}
 
+	ownerTags := map[string]string{
+		clusterIDTagPrefix + a.ClusterID(): resourceLifecycleOwned,
+		kubernetesCreatorTag:               a.controllerID,
+	}
+
 	for _, asg := range a.autoScalingGroups {
 		// This call is idempotent and safe to execute every time
-		if err := updateTargetGroupsForAutoScalingGroup(a.autoscaling, targetGroupARNs, asg.name); err != nil {
+		if err := updateTargetGroupsForAutoScalingGroup(a.autoscaling, a.elbv2, targetGroupARNs, asg.name, ownerTags); err != nil {
 			log.Printf("UpdateTargetGroupsAndAutoScalingGroups() failed to attach target groups to ASG: %v", err)
 		}
 	}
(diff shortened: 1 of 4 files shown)


56quarters:
@mikkeloscar awesome! I'll test today

56quarters:
I tested running two instances of this version of the controller in our cluster. They both seemed to only attach and detach target groups that they owned. Thanks!

LGTM

mikkeloscar:
@56quarters thanks for testing!! (since I didn't have time myself).

mikkeloscar:
:+1:

56quarters:
Actually, I might have spoke too soon. While this fixes the issue of un-owned target groups being removed - the controller seems to repeatedly fail to fetch tags for target groups attached to a load balancer when a new CF stack is first created. I'll see if I can troubleshoot further.

56quarters:
Confirming, when the ALB is first created it has no target groups and thus the array of ARNs given to the call to [`DescribeTags()`](https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/177/files#diff-250223dfa6776b8397a575de5e3efa75R106) is empty and fails. This short circuits the method which prevents any target groups from being attached to the ALB.

Steps to reproduce:
* Start with a cluster running this version of the controller with no ingress rules or CF stacks
* Create a new ingress rule
* The controller creates a CF stack including an ALB and target group
* The `updateTargetGroupsForAutoScalingGroup` method attempts to get tags for all target groups attached to the ALB
* There are no target groups attached to the ALB so the `DescribeTags()` call fails
* The code to attach target groups to the ALB never runs
* This repeats on each loop of the controller

mikkeloscar:
@56quarters Thanks for testing! 

Mind trying the new image?: `registry.opensource.zalan.do/teapot/kube-ingress-aws-controller-test:pr-177-2` :)

56quarters:
New image seems to be working perfectly - it was able to create new CF stacks from scratch and two instances are only managing their own target groups. Thanks for all your work on this!

mikkeloscar:
> Thanks for all your work on this!

Thank you!

mikkeloscar:
:+1:

szuecs:
:+1:

PR #183, opened 2018-06-29
Rename managedItem to loadBalancer

mikkeloscar:
This is a purely cosmetic change to make the code more understandable. The `managedItem` name was always confusing. It really describes a load balancer, so let's call it that!


mikkeloscar:
:+1:

hjacobs:
:+1:

PR #184, opened 2018-06-29
Fix handling limit of certificates per ALB

mikkeloscar:
This fixes a couple of bugs related to exceeding the maximum number of
certificates per ALB.

It limits the max size to 24 instead of 25. This is done because we need
to duplicate the default certificate to work around a CloudFormation bug
(#162) and therefore need one extra space for this limiting the maximum
unique certificates per ALB to 24 instead of the AWS limit of 25.

It also fixes a bug in `AddIngress` which could potentially add some of
the certificates for a single ingress to a stack and the rest to
another stack resulting in an undesired state.

Thirdly it adds rollback complete states to `IsComplete()` to
automatically attempt to update stacks that are in a rollback complete
state.

Lastly it removes some of the getter methods from structs and instead
exports the fields that are needed in other packages. This was done to
make it easier to use the structs in tests and because we are not
writing Java.

Also switched from glide to dep for vendoring because I can't be dealing with glide at this point :)

Fix #176, #175, #98

PR #187, opened 2018-07-03
Fix certificates ARNs per ALB when several are needed

### LINKED ISSUE #175: When multiple ALBs, not attaching certificates on new ones
Hi,

We're facing an issue when more than 25 certificates are needed.

When the limit (done by AWS) is reached, a new ALB is created (this is done correctly). The problem is that when a new certificate has to be added to the second ALB and update the ingress status, this is not happening.

On new domains, the ingress status is set to the first ALB, but it should be to the second one (after attaching the certificate).

I don't know exactly where this happens on the code. I'll try to search and point it here.

Thanks.
KR.

ese:
This fix #175 

## What expect

When you have a number of ingresses that leads to use more than 25 certificates, new stacks are created in order to distribute the certificates across them with a max of 25 in each of them

## What happens

The controller try to attach all of the certifacates to the first stack which leads to failure in cloudformation.
Other stacks are also created with their specific certificates, it seems that the bug is happening in only one of them

### DIFF
--- a/worker.go
+++ b/worker.go
@@ -109,7 +109,7 @@ func (l *loadBalancer) CertificateARNs() map[string]time.Time {
 	}
 
 	for arn, ttl := range l.stack.CertificateARNs() {
-		if _, ok := certificates[arn]; !ok {
+		if _, ok := certificates[arn]; ok {
 			if ttl.IsZero() {
 				certificates[arn] = time.Now().UTC().Add(l.certTTL)
 			} else if ttl.After(time.Now().UTC()) {


szuecs:
As I already explained this will not fix the issue, it will make it worse.

In https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/184#issuecomment-402424696 I added a comment, that I tested the proposed solution from @mikkeloscar, which works correctly separating 27 ingress objects with 27 certificates (one each separated) and KIAC splits it into 2 stacks, one with 24 and the other with 3 certificates.

I understand, that the issue is bad and you want to get it fixed, but please understand that switching the condition has not the effect you like to have. I tried to explain it in this comment https://github.com/zalando-incubator/kube-ingress-aws-controller/issues/175#issuecomment-400063926.

ese:
ok thanks @szuecs Im closing this becouse I am not sure about the change, and try to make more deeply details in #184

PR #190, opened 2018-07-10
Pass target port on stack update

### LINKED ISSUE #189: Target Port is always setted with default value
version: v0.7.1

Even using `-target-port=9989` the target groups are being configured to with 9999 as backend port

This was previously working setting `-health-check-port=9989`

mikkeloscar:
Passes the target port parameter on stack update to ensure it doesn't
get reset to the default value in case a custom value was set by the
user.

Fix #189


hjacobs:
:+1:

mikkeloscar:
:+1:

Issue #191, opened 2018-07-19
log how many certificates were found


szuecs:
We currently know how many EC2 and ASG and ... we have, but we lack the number of certificates we found in the logs.

```
2018/07/20 09:53:03 Start polling sleep 30s
2018/07/20 09:53:33 Found 50 ingress(es)
2018/07/20 09:53:33 Found 1 stack(s)
2018/07/20 09:53:34 Found 4 auto scaling group(s)
2018/07/20 09:53:34 Found 0 single instance(s)
2018/07/20 09:53:34 Found 19 EC2 instance(s)
2018/07/20 09:53:34 Failed to find certificate for hostname <name>
..
2018/07/20 09:53:34 Have 1 model(s)
2018/07/20 09:53:34 Ingress update not needed <ns>/<name> with DNS name "aws-0856-lb-....eu-central-1.elb.amazonaws.com"
... a lot of messages which we might want to move to debug logging
```

so0k:
Currently the logs also lack status levels, which is a bit annoying to handle in centralised log system, adopting glog (as it's a standard across all k8s projects) might be something good (although that might need to go under a separate issue)

szuecs:
@so0k We will not change it to glog, because the agreement across the Go community is that, it's shit, because it makes it impossible to use something else then the flags package or use a package as library.

For the status levels we could change the logging to logrus.

mikkeloscar:
Related to #89

so0k:
@mikkeloscar - thank you - I searched for log related issue but didn't found that one, sorry for putting my comment in wrong issue

@szuecs - got it 😆

mikkeloscar:
Closed in #192

PR #192, opened 2018-07-20
show number of certs found

szuecs:
#191


mikkeloscar:
We could probably optimize this further to not do multiple certificate lookups on each run, but this can be done separately.

mikkeloscar:
:+1:

PR #194, opened 2018-07-22
Cleanup controller initialization and termination

mikkeloscar:
This is just a bit of cleanup around the control loop. Mostly cosmetic.

* Run `doWork` in the beginning of the control loop instead of always waiting 30s before doing any work.
* Use a `context` for stopping the control loop and get rid of the custom channels.


szuecs:
:+1:

PR #198, opened 2018-07-26
Detach non-existing target groups from ASGs

### LINKED ISSUE #197: Target groups not being removed from ASG if stack is deleted
In case a Load Balancer stack is deleted, the Target group will also be deleted but it will still be defined in the autoscaling group, preventing AWS from creating new instances in the ASG.

> Launching a new EC2 instance: i-abc. Status Reason: Failed to update target group arn:aws:elasticloadbalancing:eu-central-1:123456:targetgroup/kube-ing-TG-Q1NIZBFGOEXH/277d10c4b9018dbc: Target group 'arn:aws:elasticloadbalancing:eu-central-1:123456:targetgroup/kube-ing-TG-Q1NIZBFGOEXH/277d10c4b9018dbc' not found Updating load balancer configuration failed

The problem is that we only list the TGs and loop through that list to find those that should no longer be attached to the ASG. We also need to check what is in the ASG but no longer in the TG list. https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/4b747629dd09fc9a924b95926bf3705bdf9330c7/aws/asg.go#L92

mikkeloscar:
In case an Ingress stack is deleted, the TargetGroup will also be
deleted, but not removed from the list of TargetGroups in the ASG.

This renders the ASG in a state where it can't spawn new instances
because it can't attach the non-existing TargetGroup.

To fix this, the controller will now look up all TargetGroups every time
it consider TargetGroups to attachment/detachment of an ASG.

Fix #197

### DIFF
--- a/aws/asg.go
+++ b/aws/asg.go
@@ -1,6 +1,7 @@
 package aws
 
 import (
+	"context"
 	"fmt"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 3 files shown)


mikkeloscar:
:+1:

PR #205, opened 2018-08-24
Allow control of the SSL policy applied to https listeners via flag

jhohertz:
`-ssl-policy=<policyname>`. Currently defaults to the same AWS
default of 'ELBSecurityPolicy-2016-08' which is less secure cipher/
protocol wise than some of the others.

See: https://docs.aws.amazon.com/elasticloadbalancing/latest/application/create-https-listener.html#describe-ssl-policies

(Hello again!)


hjacobs:
Thanks for the PR! Good find, I guess we should also use a more strict SSL policy in Zalando by default..

hjacobs:
:+1:

szuecs:
👍

hjacobs:
@jhohertz thanks! :smile:

PR #212, opened 2018-09-05
Ignore certificates that can't be verified

### LINKED ISSUE #211: Cert matching can select non trusted certificates
We should support non trusted (self signed) certificates, but if we have similar matching results we should not just select the newer one, but check if both are trusted first and pick the trusted before checking "newer".

aermakov-zalando:
* Fetch the certificate and chain information from IAM/ACM
 * Verify validity of the certificate using system roots (for TLS usage)  and skip invalid certificates automatically

Fix #211.


szuecs:
:-1:

szuecs:
Please don't change the logic, but only add another test of the cert if you have more than one valid one.
Check first the best match, than the CA validity, after that pick the newest.
If you do it different we will run into problems.

aermakov-zalando:
See the comment in #211. There's no point in having unverifiable and verifiable certificates on the same ELB (I'm not even sure if the cert selection algorithm is documented or standardised). You'll have the same issues where one of the clients will randomly not be able to verify the certificate.

If you really want to support this use case, I'd rather create a separate PR that'll only allow unverifiable certificates by explicitly specifying the ARN (and force a separate ELB in this case). However, is it really needed for now?

szuecs:
@aermakov-zalando yes separate PR makes sense, but don't change the cert matching rule, but rather filter invalid certs before applying the cert matching.
Thanks!

aermakov-zalando:
@szuecs The only change in matching logic is that unverifiable certificates are completely ignored, what exactly do you suggest to change?

aws/acm.go
```diff
@@ -49,14 +51,24 @@ func getACMCertificateSummaries(api acmiface.ACMAPI) ([]*acm.CertificateSummary,
 }
 
 func getCertificateSummaryFromACM(api acmiface.ACMAPI, arn *string) (*certs.CertificateSummary, error) {
-	params := &acm.DescribeCertificateInput{CertificateArn: arn}
-	resp, err := api.DescribeCertificate(params)
+	params := &acm.GetCertificateInput{CertificateArn: arn}
+	resp, err := api.GetCertificate(params)
```
szuecs: Ca nyou update https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/master/deploy/requirements.md , we do not have the GetCertificate allowed and please add a readme entry how to switch the version, because the iam role needs to be changed, so we also have to increase the version number as before.

szuecs:
Then I miss read your changes, sorry had not too much time looking into this.
I checked it now and it looks good but we need to add the documentation as I commented

szuecs:
:+1:

mikkeloscar:
:+1:

PR #221, opened 2018-10-23
fix #220, use correct controllerID looking up SGs

### LINKED ISSUE #220: Logic error when searching for security group
This line:

https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/master/aws/ec2.go#L305

Is doing the wrong thing, in that it only looks for the `DefaultControllerID` in the value of the `kubernetes:application` tags, where it should be using the specified controller ID if set.

Planning to create a PR for this.

jhohertz:
This went a little deeper than initially expected, owing to the fact that where buildManifest was being called (and where the SGs are ultimately looked up), was *prior* to the application of the collected flag values to the awsAdapter.

So I've moved the manifest to something similar to the With*() functions, and apply it in the same way *after* we have the flags applied to the awsAdapter. Let me know if you'd like to see this handled differently?

I had not previously seen this, as I had always been running an instance with the default controllerID (and so had a SG of that name around to be successfully looked up). It's only recently that has been absent, and exposed this bug.

### DIFF
--- a/aws/adapter.go
+++ b/aws/adapter.go
@@ -130,7 +130,7 @@ func defaultConfigProvider() client.ConfigProvider {
 // Before returning there is a discovery process for VPC and EC2 details. It tries to find the Auto Scaling Group and
 // Security Group that should be used for newly created Load Balancers. If any of those critical steps fail
 // an appropriate error is returned.
-func NewAdapter() (adapter *Adapter, err error) {
+func NewAdapter(newControllerID string) (adapter *Adapter, err error) {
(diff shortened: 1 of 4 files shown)


szuecs:
:+1:

jhohertz:
Yes, that does look better I think too. Thanks @mikkeloscar

mikkeloscar:
:+1:

szuecs:
:+1:

szuecs:
Thanks @jhohertz

Issue #201, opened 2018-08-10
Access logs configuration


sgmiller:
It'd be nice to be able to configure the Access Logs for the ALBs created.  Even better would be to have different logging per host, but that's impossible without a separate ALB per domain.

szuecs:
If you would like to add a PR would be fine!

szuecs:
closed by https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/222

PR #222, opened 2018-10-25
Enable ALB logging to S3 via new flags

jhohertz:
Two new flags have been implemented:

  --logs-s3-bucket <name of s3 bucket>
  --logs-s3-prefix <path prefix within s3 bucket>

Setting the bucket will enable logging support, not setting it will
leave it disabled. The prefix setting will only take effect if the
bucket is also set.

This should satisfy the feature request in #201

### DIFF
--- a/aws/cf_template.go
+++ b/aws/cf_template.go
@@ -22,7 +22,7 @@ func hashARNs(certARNs []string) []byte {
 	return hash.Sum(nil)
 }
 
-func generateTemplate(certs map[string]time.Time, idleConnectionTimeoutSeconds uint) (string, error) {
+func generateTemplate(certs map[string]time.Time, idleConnectionTimeoutSeconds uint, albLogsS3Bucket, albLogsS3Prefix string) (string, error) {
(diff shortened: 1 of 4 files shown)


aws/cf_template.go
```diff
@@ -22,7 +22,7 @@ func hashARNs(certARNs []string) []byte {
 	return hash.Sum(nil)
 }
 
-func generateTemplate(certs map[string]time.Time, idleConnectionTimeoutSeconds uint) (string, error) {
+func generateTemplate(certs map[string]time.Time, idleConnectionTimeoutSeconds uint, albLogsS3Bucket string, albLogsS3Prefix string) (string, error) {
```
szuecs: nitpick: you can use `func f(a,b,c string) {}` in Go.

szuecs:
:+1:

jhohertz:
Thanks for the feedback @szuecs, addressed the nit you raised.

mikkeloscar:
:+1:

hjacobs:
:+1:

Issue #206, opened 2018-08-30
Split shared security groups of ALB-internal and ALB-external


marandalucas:
### kube-ingress-aws-controller VERSION
```kube-ingress-aws-controller v0.7.2```

### Kubernetes version
```
Client Version: v1.9.3
Server Version: v1.9.6
```

### Environment
- **Cloud provider or hardware configuration**: AWS with Kops 1.9.6 + RBAC authorization

### What happened
When for the first time I create ingress resources with internal annotations (Internal ALB) or no annotations (Public ALB), the cloudformation provisions the two ALBs, but it takes the same security groups for them.

### Expected behavior
I have to apply differents restrictions depends if the ALB is internal or public.
Would be possible to split the shared security groups of ALB-internal and ALB-external?

### Workaround
I deleted the automatic security groups from one ALB and apply new security groups with our necessity. But if it delete, i need to update with the last change.
Could I have a problem in the future with this manual change?

Issue #23, opened 2017-02-09
Consider allowing to configure ALB security group


hjacobs:
Idea: allow configuring the ALB security group via annotations:

* the optional `zalando.org/aws-load-balancer-security-group` annotation can select a different AWS Security Group by name
* the default AWS Security Group can be configured via a command line option

This allows using a default security group with IP restrictions and explicitly exposing the ALB by selecting an unrestricted security group.

lmineiro:
Could we consider another, similar, annotation where a ruleset could be defined?
I'm thinking about those scenarios where a deployer user doesn't have the access level that allows him to create a security group to reference but also wants to customize the security rules for his load balancer.

hjacobs:
@lmineiro I would try to keep it simple, specifying the full set of ingress rules is overkill (as annotation). My scenario does not necessarily require AWS rights, the security groups could pre-provisioned for known use cases (something like "restricted/company-internal" SG with IP restrictions and another one "public" for exposure).

lmineiro:
Fair enough. Let's keep it simple

joberdick:
@hjacobs any thoughts on specifying security group by a tag instead of SG name? We automatically generate clusters with supporting SG's. Our SG names are dynamic. We could easily tag them with consistent names such as "internet-public".

szuecs:
@joberdick feel free to create a new issue for your use case. 
The question from my point of view is:
- How this should work for shared ALBs and the user point of view?
- Would you tag all ingress with the same SG-Tag annotation? 

If you can write down what you want and define a workflow how to use it, it would be simpler to think about the solution.

PR #230, opened 2018-12-13
Security group annotation support 

marcinzaremba:
Fixes https://github.com/zalando-incubator/kube-ingress-aws-controller/issues/206
Fixes https://github.com/zalando-incubator/kube-ingress-aws-controller/issues/23

### DIFF
--- a/kubernetes/adapter.go
+++ b/kubernetes/adapter.go
@@ -104,7 +104,7 @@ func (a *Adapter) newIngressFromKube(kubeIngress *ingress) *Ingress {
 	}
 }
 
-func (a *Adapter) newIngressForKube(i *Ingress) *ingress {
+func newIngressForKube(i *Ingress) *ingress {
 	shared := "true"
 
 	if !i.Shared {
@@ -172,5 +172,5 @@ func (a *Adapter) UpdateIngressLoadBalancer(ingress *Ingress, loadBalancerDNSNam
 		return ErrInvalidIngressUpdateParams
 	}
 
-	return updateIngressLoadBalancer(a.kubeClient, a.newIngressForKube(ingress), loadBalancerDNSName)
+	return updateIngressLoadBalancer(a.kubeClient, newIngressForKube(ingress), loadBalancerDNSName)
 }
--- a/kubernetes/adapter.go
+++ b/kubernetes/adapter.go
@@ -9,9 +9,9 @@ import (
 )
 
 type Adapter struct {
-	kubeClient     client
-	ingressFilters []string
-	securityGroup  string
+	kubeClient                  client
+	ingressFilters              []string
+	ingressDefaultSecurityGroup string
 }
 
 var (
@@ -52,7 +52,7 @@ func (i *Ingress) String() string {
 }
 
 // NewAdapter creates an Adapter for Kubernetes using a given configuration.
-func NewAdapter(config *Config, ingressClassFilters []string, securityGroup string) (*Adapter, error) {
+func NewAdapter(config *Config, ingressClassFilters []string, ingressDefaultSecurityGroup string) (*Adapter, error) {
 	if config == nil || config.BaseURL == "" {
 		return nil, ErrInvalidConfiguration
 	}
@@ -60,7 +60,7 @@ func NewAdapter(config *Config, ingressClassFilters []string, securityGroup stri
 	if err != nil {
 		return nil, err
 	}
-	return &Adapter{kubeClient: c, ingressFilters: ingressClassFilters, securityGroup: securityGroup}, nil
+	return &Adapter{kubeClient: c, ingressFilters: ingressClassFilters, ingressDefaultSecurityGroup: ingressDefaultSecurityGroup}, nil
 }
 
 func (a *Adapter) newIngressFromKube(kubeIngress *ingress) *Ingress {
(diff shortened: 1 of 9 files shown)


kubernetes/adapter.go
```diff
@@ -86,10 +100,11 @@ func newIngressFromKube(kubeIngress *ingress) *Ingress {
 		Scheme:         scheme,
 		Hostnames:      hostnames,
 		Shared:         shared,
+		SecurityGroup:  kubeIngress.getAnnotationsString(ingressSecurityGroupAnnotation, a.ingressDefaultSecurityGroup),
 	}
 }
 
-func newIngressForKube(i *Ingress) *ingress {
+func (a *Adapter) newIngressForKube(i *Ingress) *ingress {
```
szuecs: you don't need an instance of Adapter here

kubernetes/adapter_test.go
```diff
@@ -27,6 +31,7 @@ func TestMappingRoundtrip(tt *testing.T) {
 				CertificateARN: "zbr",
 				Shared:         true,
 				Hostnames:      []string{"domain.example.org"},
+				SecurityGroup:  "sg-123456",
```
szuecs: you can define this as var and reuse in tests

kubernetes/adapter_test.go
```diff
@@ -64,6 +70,7 @@ func TestMappingRoundtrip(tt *testing.T) {
 				Scheme:         "internal",
 				CertificateARN: "zbr",
 				Shared:         false,
+				SecurityGroup:  "deadbeef",
```
szuecs: you can define this as var and reuse it

kubernetes/adapter_test.go
```diff
@@ -87,25 +95,18 @@ func TestMappingRoundtrip(tt *testing.T) {
...
-			}
-			if got.CertificateARN != tc.kubeIngress.Metadata.Annotations[ingressCertificateARNAnnotation] {
-				t.Error("wrong value from CertificateARN()")
-			}
-			if got.Scheme != tc.kubeIngress.Metadata.Annotations[ingressSchemeAnnotation] {
-				t.Error("wrong value from Scheme()")
-			}
-			if got.Hostname != tc.kubeIngress.Status.LoadBalancer.Ingress[1].Hostname {
-				t.Error("wrong value from Hostname()")
+				t.Errorf("mapping from kubernetes ingress to adapter failed. wanted %#v, got %#v", tc.ingress, got)
```
szuecs: Maybe also change the message to use cmp.Diff(test.before, test.after) from https://godoc.org/github.com/google/go-cmp/cmp#example-Diff--Testing

worker.go
```diff
@@ -266,25 +275,33 @@ func buildManagedModel(certs []*certs.CertificateSummary, certsPerALB int, certT
...
-		// non-matching scheme or too many certificates, add a new
+		// non-matching scheme, non-matching security group or too many certificates, add a new
 		// stack.
 		if !added {
 			i := make(map[string][]*kubernetes.Ingress, len(certificateARNs))
 			for _, certificateARN := range certificateARNs {
 				i[certificateARN] = []*kubernetes.Ingress{ingress}
 			}
-			model = append(model, &loadBalancer{ingresses: i, scheme: ingress.Scheme, shared: ingress.Shared})
+			loadBalancers = append(loadBalancers, &loadBalancer{ingresses: i, scheme: ingress.Scheme, shared: ingress.Shared, securityGroup: ingress.SecurityGroup})
```
szuecs: please do this multiline, it's already quite long

worker.go
```diff
@@ -266,25 +275,33 @@ func buildManagedModel(certs []*certs.CertificateSummary, certsPerALB int, certT
...
 		}
 	}
 
+	return loadBalancers
+}
+
+func buildManagedModel(certs []*certs.CertificateSummary, certsPerALB int, certTTL time.Duration, ingresses []*kubernetes.Ingress, stacks []*aws.Stack) []*loadBalancer {
+	sortStacks(stacks)
+	model := getAllLoadBalancers(certTTL, stacks)
+	model = matchIngressesToLoadBalancers(model, certs, certsPerALB, ingresses)
```
szuecs: You don't need really to assign the value `model` if you change the var in matchIngressesToLoadBalancers that you pass, because a slice is reference value and we have pointers as values inside. 
If you think it's more readable like this, then it's fine to just let it like this.
Comment is more FYI ;)

marcinzaremba: Thanks ;) In my opinion it's just more readable and looks exactly how `append` look like and what actually is done here is really appending to the slice under some conditions.

marcinzaremba:
👍

kubernetes/adapter_test.go
```diff
@@ -87,31 +97,16 @@ func TestMappingRoundtrip(tt *testing.T) {
...
-			if got.Scheme != tc.kubeIngress.Metadata.Annotations[ingressSchemeAnnotation] {
-				t.Error("wrong value from Scheme()")
-			}
-			if got.Hostname != tc.kubeIngress.Status.LoadBalancer.Ingress[1].Hostname {
-				t.Error("wrong value from Hostname()")
-			}
-			if got.String() != fmt.Sprintf("%s/%s", tc.ingress.Namespace, tc.ingress.Name) {
-				t.Error("wrong value from String()")
-			}
+			a, _ := NewAdapter(testConfig, testIngressFilter, testIngressDefaultSecurityGroup)
```
szuecs: ignore the returned error?

szuecs:
:+1:

marcinzaremba:
👍

szuecs:
:+1:

PR #232, opened 2018-12-18
Add more unit tests

marcinzaremba:
Part of increasing test coverage initiative.

### DIFF
--- a/worker.go
+++ b/worker.go
@@ -149,6 +149,45 @@ func (l *loadBalancer) Owner() string {
 	return ""
 }
 
+// CertificatesFinder interface represents a list of certificates
+// and some basic operations than can be performed on them.
+type CertificatesFinder interface {
+	CertificateSummaries() []*certs.CertificateSummary
+	CertificateExists(certificateARN string) bool
+	FindMatchingCertificateIDs([]string) []string
+}
+
+// Certificates represents a generic list of certificates
+type Certificates struct {
+	certificateSummaries []*certs.CertificateSummary
+}
+
+// CertificateSummaries returns summaries of all certificates
+func (c *Certificates) CertificateSummaries() []*certs.CertificateSummary {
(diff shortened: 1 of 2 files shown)


szuecs:
:+1:

worker.go
```diff
@@ -149,6 +149,45 @@ func (l *loadBalancer) Owner() string {
 	return ""
 }
 
+// CertificatesFinder interface represents a list of certificates
+// and some basic operations than can be performed on them.
+type CertificatesFinder interface {
```
mikkeloscar: @marcinzaremba since you refactor this, would it make sense to move it out of `worker.go` to a separate file? `worker.go` is already a bit bloated with stuff :)

marcinzaremba: I was thinking about this but cannot come up with meaningful separate component (area of responsibility. And I would like not to end up with something pretty meaningless like `worker_utils.go`. Can you maybe suggest some possibilities?

mikkeloscar: `certificates_finder.go` would be fine for me, it doesn't have to be in a separate package, just to things out of `worker.go`. If you don't want to do it in this PR, it's also totally fine, I just wanted to suggest it :)

szuecs:
:+1:

szuecs:
@mikkeloscar I think it’s fine like this, merge?

marcinzaremba:
Initially I wanted to also test `matchIngressesToLoadBalancers` as I believe this the most critical part of functionality here (with edge cases like out of certificates for LB, shared LB, now security group matching, etc.). I believe this little refactor enables this (provide a mock with interface of `CertificateFinder` to `matchIngressesToLoadBalancers`. So we can merge it in the state as is and let testing this function for others.

marcinzaremba:
👍

szuecs:
Thanks @marcinzaremba great work!

Issue #164, opened 2018-05-21
support for AWS WAF


sstarcher:
I would be great to be able to configure a AWS WAF in front of the ALB.

szuecs:
Hi!
Makes sense to me, but can you please specify what exactly you want to specify and how this should be configured?
I never used it myself. 

For example ingress annotation `aws-waf: <name>`
or `aws-waf-regexp: <re2-regexp>`

sstarcher:
For my use-case `aws-waf: <name>` would be sufficient and a great addition

greenboxal:
:+1:

@szuecs Something related that still bugs me, how do you guys keep in sync resources that need to be created outside K8S?

I have this problems with security groups, IAM roles (for use with kube2iam), S3 buckets and other resources. Today I create those extra resources with Terraform and "hope" that name defined on Kubernetes annotations are the same.

If WAF support is added we would have the same problem.

szuecs:
@greenboxal it depends, but most of the time the solution is to crash and let kubelet restart your container.
SG are really nasty, because AWS does not send a RST packet, from network point of view I consider this as bug, but most network/firewall engineers I know think this as a security feature to drop silently a packet (security by obscurity in my point of view).
kube2iam specific: use init container in your pod to validate that the node already got the iam role, if it does not have it crash the init container.

For WAF I would not care so much. We would only configure WAF if you set a rule or name in the annotation and let cloudformation figure dependenices out. What do you think can hurt if WAF rule is not yet applied and you have an ALB or do you have any other scenario that might be a problem?

janario:
It would be great to have that, we have steps that have to be executed right after the ALB creating, in order to keep it secure

I've sent a PR https://github.com/zalando-incubator/kube-ingress-aws-controller/issues/164

Please take a look :)

szuecs:
@janario thanks for the PR. Your implementation does a fixed WAF rule, that is passed via CLI flag, which is fine and a step forward. 
Do you want to also have one with ingress annotation with opt-in or opt-out of the fixed WAF from CLI params to the controller?

szuecs:
I assume that, you need to force to create another ALB.
We already have a similar annotation `zalando.org/aws-load-balancer-shared: "false"` to force a non shared ALB. I see 3 solutions for this:
[...]

PR #237, opened 2019-01-24
#164 Added impl that provide waf id and create cf with waf association

janario:
Added parameter `aws-waf-web-acl-id` (it must be the waf hash id, not the name) that when it is provided it will create an cludformation template with an WafAssociation

It will require the policies:

- waf-regional:AssociateWebACL
- waf-regional:DisassociateWebACL
- elasticloadbalancing:SetWebACL

Known issue, It won't disassociate in case of removal to any wafid

### DIFF
--- a/aws/adapter.go
+++ b/aws/adapter.go
@@ -58,6 +58,7 @@ type Adapter struct {
 	ipAddressType              string
 	albLogsS3Bucket            string
 	albLogsS3Prefix            string
+	wafWebAclId                string
 }
 
 type manifest struct {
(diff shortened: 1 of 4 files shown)


mikkeloscar:
@janario thanks for the PR, please sign-off your commit! (See DCO check).

szuecs:
:+1:

szuecs:
We try to get ` CDP/pr/build Expected — Waiting for status to be reported ` optional to enable merging forks...

mikkeloscar:
:+1:

Issue #236, opened 2019-01-24
NLB Support (AWS Launched TLS Termination for NLB)


hjacobs:
AWS Network Load Balancer (NLB) now supports TLS termination, so we might want to get rid of ALB: https://aws.amazon.com/blogs/aws/new-tls-termination-for-network-load-balancers/

Passpes:
is this feature added already?

mikkeloscar:
No, not yet!

so0k:
We've been using NLB with TLS termination for our clusters (managed by Terraform for now). Interesting to note no SNI support, so will need to provision an NLB for each cert that matches the ingress host requests.

joberdick:
@so0k This seems impractical without SNI support. You can only have a few NLBs in an account. Bummer. I was interested in this.

so0k:
We switched back to Cert Manager and TLS termination at our ingress proxy due to some WSS issues. It’s been crazy, but we see the light at the end of the tunnel

szuecs:
@so0k did you used kube-ingress-aws-controller + skipper to do websockets (WSS?)?
If so we introduced a bug in skipper with a loadbalancer refactoring that was fixed in https://github.com/zalando/skipper/releases/tag/v0.10.186.

andreasph:
@joberdick since a little more than a month NLB has SNI support (https://aws.amazon.com/about-aws/whats-new/2019/09/elastic-load-balancing-network-load-balancers-now-supports-multiple-tls-certificates-using-server-name-indication/)

I'm currently using kube-ingress-aws-controller + ingress-nginx (no skipper) and I have a problem with http2 requests forwarded to the targets by the ALB as http1.1: this prevents us to use it for grpc services

szuecs:
@andreasph just to make sure the basics are met: Do you have a https target group (`ALB -> nginx`)?

andreasph:
@szuecs the whole setup is working fine with http ingress. gprc is supported as well by nginx ingress, the problem is that the external HTTP2 grpc request is then forwarded by the ALB to the target group as HTTP/1.1

it is somehow similar to this problem: https://github.com/kubernetes/ingress-nginx/issues/2963#issuecomment-415141659

does skipper support grpc ingress?
I preferred to use nginx ingress rather than skipper for broader support and easier troubleshooting

andreasph:
[...]

PR #242, opened 2019-03-16
Add support for Network Load Balancers

szuecs:
first part to support NLB #236

resource: 
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticloadbalancingv2-loadbalancer.html 

Signed-off-by: Sandor Szuecs <sandor.szuecs@zalando.de>

### DIFF
--- a/controller.go
+++ b/controller.go
@@ -65,6 +65,7 @@ var (
 	firstRun                   bool = true
 	cwAlarmConfigMap           string
 	cwAlarmConfigMapLocation   *kubernetes.ResourceLocation
+	loadBalancerType           string
 )
 
 func loadSettings() error {
@@ -112,6 +113,7 @@ func loadSettings() error {
 	flag.StringVar(&wafWebAclId, "aws-waf-web-acl-id", aws.DefaultWafWebAclId, "Waf web acl id to be associated with the ALB")
 	flag.StringVar(&cwAlarmConfigMap, "cloudwatch-alarms-config-map", "", "ConfigMap location of the form 'namespace/config-map-name' where to read CloudWatch Alarm configuration from. Ignored if empty.")
 	flag.BoolVar(&httpRedirectToHttps, "redirect-http-to-https", defaultHttpRedirectToHttps, "Configure HTTP listener to redirect to HTTPS")
+	flag.StringVar(&loadBalancerType, "load-balancer-type", aws.DefaultLoadBalancerType, "Sets default Load Balancer type (application or network).")
 
 	flag.Parse()
 
@@ -166,6 +168,10 @@ func loadSettings() error {
 		}
 
 		cwAlarmConfigMapLocation = loc
+	}
+
+	if loadBalancerType != "application" && loadBalancerType != "network" {
+		return fmt.Errorf("invalid load-balancer-type, must be 'application' or 'network'")
 	}
 
 	if quietFlag && debugFlag {
@@ -277,7 +283,7 @@ func main() {
 		ingressClassFiltersList = strings.Split(ingressClassFilters, ",")
 	}
 
-	kubeAdapter, err = kubernetes.NewAdapter(kubeConfig, ingressClassFiltersList, awsAdapter.SecurityGroupID(), sslPolicy)
+	kubeAdapter, err = kubernetes.NewAdapter(kubeConfig, ingressClassFiltersList, awsAdapter.SecurityGroupID(), sslPolicy, loadBalancerType)
 	if err != nil {
 		log.Fatal(err)
 	}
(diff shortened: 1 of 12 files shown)


so0k:
we run a lot of NLB without TLS now (we use Cert Manager and do TLS Termination by Envoy proxy)

mikkeloscar:
@szuecs FYI: I rebased this on master

szuecs:
Thanks @mikkeloscar

mikkeloscar:
@szuecs I now also added per ingress configuration for selecting load balancer type and tested in our cluster to validate the upgrade path.

(Needed to do some coding :D)

It needs more testing. I didn't try creating network load Balancers yet. And I think we also need to only conditionally enable HTTPS redirect if `type=application`.

mikkeloscar:
And ofc. docs are also missing :)

mikkeloscar:
Also securityGroups can't be set with `network` load balancer

```
Security groups are not supported for load balancers with type 'network' (Service: AmazonElasticLoadBalancingV2; Status Code: 400; Error Code: InvalidConfigurationRequest; Request ID: 04ded6d8-c1c7-45ea-9dd0-cf483f761700)
--
```

mikkeloscar:
* Updated to handle all cases where some configuration can only be set for one type of load balancer: `network` or `application`
* Successfully deployed a working NLB for an ingress in our test cluster by specifying `zalando.org/aws-load-balancer-type: network`

I found one interesting setting for NLBs: [cross-zone-load-balancing](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/network-load-balancers.html#cross-zone-load-balancing) which when enabled let's NLB load balance across zones. (This is the default and only option for ALBs, but it's configurable for NLBs).

What's interesting is how this setting behaves for the NLB. If you deploy the NLB spanning 3 AZs then you by default get 3 IPs when you resolve the NLB hostname. However if the TargetGroup of the NLB only has healthy instances in 2/3 AZs then you get only 2 IPs resolved. This means that it should be possible to not have cross-zone-load-balancing thus presumably reducing the latency (and traffic cost?). This I have not tested.
However, I worry that updating the DNS record could take too long so if you loose an instance in one AZ, you could still have clients going to the IP of that AZ but the NLB having nowhere to route. This is not a problem in a cross-zone-enabled setup because all NLB interfaces can route to all instances no matter the AZ. We have to test what impact it would have in our Zalando setup. For now I have added a flag on the controller which enables/disables cross-zone-load-balancing for all `network` load balancers. It's disabled by default for now.

Another question is what values we should use for the `zalando.org/aws-load-balancer-type` annotation. Right now we use `application` and `network` because this is the term AWS uses. However service type `LoadBalancer` use the term `nlb` and `elb`: https://github.com/kubernetes/kubernetes/blob/2905275a08f8e3999e7e6bc8b02f407fd6a0ff36/staging/src/k8s.io/legacy-cloud-providers/aws/aws_loadbalancer.go#L66 so maybe we want to call it `alb` and `nlb` to make it more similar for users. WDYT?

## TODO

* [ ] More testing
* [ ] Documenting the feature

controller.go
```diff
@@ -65,6 +65,7 @@ var (
 	firstRun                   bool = true
 	cwAlarmConfigMap           string
 	cwAlarmConfigMapLocation   *kubernetes.ResourceLocation
+	loadBalancerType           string
```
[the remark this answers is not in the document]
mikkeloscar: Probably makes sense to use kingpin for flag handling at this point as it can do a lot of that out of the box. I would do that in a separate PR though. WDYT?

mikkeloscar:
@szuecs I think we can test the non-cross AZ traffic later as it's just a flag here.

One open thing for this PR is the naming in the ingress, what do you think is best? `application` and `network` as AWS uses or `alb` and `nlb` as is used for the service of type LoadBalancer? I'm mostly leaning towards the `alb`, `nlb` version because it IMO creates a more unified user experience.

szuecs:
I don’t have an opinion on that: nlb / alb seems fine to me.

mikkeloscar:
* Changed it to be `nlb`/`alb` for the annotation values.
* Documented the feature in the README
* Tested in our test cluster and found a couple of unrelated bugs which I also fixed (Some AWS API call have limits, so you can e.g. only update 10 target groups at a time.)

szuecs:
@mikkeloscar so ready to merge? Then I will drop the label and do my last review

mikkeloscar:
@szuecs yes, ready for review :)

mikkeloscar:
:+1:

worker.go
```diff
@@ -333,6 +339,14 @@ func matchIngressesToLoadBalancers(loadBalancers []*loadBalancer, certs Certific
 		// limit is exeeded.
 		added := false
 		for _, lb := range loadBalancers {
+			// TODO(mlarsen): hack to phase out old load balancers
+			// which can't be updated to include type
+			// specification.
+			// Can be removed in a later version
```
szuecs: Should we document the upgrade procedure or do we want to create these if we tested it successfully in one of our production clusters?

szuecs:
:+1:

worker.go
mikkeloscar: I would document the update procedure when we remove this comment at a later stage. Right now it works fine to upgrade (at least from my tests :))

The procedure would be to first update to the version which includes these changes and then later to a future version were this hack is removed.

szuecs:
@so0k maybe this is interesting for you

PR #244, opened 2019-04-09
Go modules

njuettner:
This PR introduces two changes:

* Go modules
* gometalinter is deprecated, instead we could try out golangci-lint


szuecs:
👍

szuecs:
thanks

njuettner:
@szuecs I've added staticcheck.
The bot also reviews the PR and makes suggestions for code changes according to what linters has been enabled, see:

image

delivery.yaml
```diff
@@ -1,15 +1,23 @@
...
 - id: build
+  overlay: ci/golang
+  cache:
+    paths:
+    - /go/pkg/mod       # pkg cache for Go modules
+    - ~/.cache/go-build # Go build cache
   type: script
   commands:
-  - desc: Build and push docker image
+  - desc: Build docker image
```
mikkeloscar: We should also run `make test` before building via CDP.

njuettner: I wasn’t sure about that because we already run make test in travis.

mikkeloscar: I would say we should have it to ensure we only build and push an image if the tests pass.

njuettner:
👍

mikkeloscar:
:+1:

mikkeloscar:
:+1:

njuettner:
👍

Issue #239, opened 2019-01-29
Controller fails to cleanup old Target groups from empty Autoscaling groups.


mikkeloscar:
Imagine the following scenario:

1. You have a cluster with two autoscaling groups **asg-a** and **asg-b**. Both ASGs have **1** instance running each.
2. You deploy an ingress resource, which triggers the controller to create an ALB and attach Target groups to the two ASGs.
3. **asg-b** is scaled down to **0** instances (for whatever reason, e.g. cluster autoscaling).
4. You delete your ingress resource and the controller decommissions the ALB and target groups.
5. The controller finds that one instance in **asg-a** and removes the "ghost" Target Group from the ASG. Because **asg-b** didn't have any instances running at this time, the controller doesn't detect that **asg-b** exists, and therefore **DOESN'T** remove the ghost target group.
6. At a later point the cluster tries to scale up **asg-b** but fails because of the ghost target group:
![image](https://user-images.githubusercontent.com/128566/51943245-4e6c1800-2419-11e9-855e-439a9767821a.png)
    ```
    Status Reason: One or more target groups not found. Validating load balancer configuration failed.
    ```
7. The cluster can no longer scale up **asg-b** until someone manually removes the ghost target group from the ASG via the AWS Console.

This bug happens because ASGs are discovered by looking at tags of running EC2 instances:

https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/c5fd0b24ae038b0dee973cfb19ae7c86e881dad6/aws/adapter.go#L622

Thus it will never discover ASGs with no instances running. We should fix this by having the controller list all the ASGs.

szuecs:
It should be fixed by the linked PR #245

PR #245, opened 2019-04-16
fetch all related autoscaling groups (#239)

gchudnov:
Hi @mikkeloscar 
unfortunately delayed the PR after the hack-night, nevertheless here it is.

Please let me know if this is the discussed solution and I'll add some tests + remove the related dead code.

Thank you,
Grigorii


gchudnov:
hm..
it looks the related file was changed:
```
<<<<<<< fetch-all-autoscaling-groups
				launchConfigurationName: aws.StringValue(g.LaunchConfigurationName),
=======
>>>>>>> master
```

not sure if this is the required fix that fixes the problem?

szuecs:
@gchudnov can you fetch origin/master, rebase on master and do a force push?

mikkeloscar:
@gchudnov thanks a lot for the PR. I'm going on easter vacation today so I will not get time to review it. But will get back to it next week. Like @szuecs suggested please rebase on master :) The change you mentioned did not fix the actual issue, so your PR is very welcome :)

gchudnov:
sure, will do it.

aws/asg.go
```diff
@@ -9,6 +9,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/autoscaling/autoscalingiface"
 	"github.com/aws/aws-sdk-go/service/elbv2"
 	"github.com/aws/aws-sdk-go/service/elbv2/elbv2iface"
+	"strings"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

mikkeloscar:
@gchudnov build is failing FYI

gchudnov:
yes, I've seen it,
will fix it soon

szuecs:
please also rebase on master :)

aws/adapter.go
```diff
@@ -636,39 +636,22 @@ func (a *Adapter) UpdateAutoScalingGroupsAndInstances() error {
...
-		if err != nil {
-			log.Errorf("failed fetching Auto Scaling Groups details: %v", err)
-		} else {
-			for name, asg := range fetchedAutoScalingGroups {
-				newAutoScalingGroups[name] = asg
-			}
+	newAutoScalingGroups := make(map[string]*autoScalingGroupDetails)
+	fetchedAutoScalingGroups, err := getOwnedAutoScalingGroups(a.autoscaling)
+	if err != nil {
+		log.Printf("failed fetching Auto Scaling Groups details: %v", err)
```
mikkeloscar: This should be `log.Errorf` now, we changed the logging in #246

aws/asg.go
```diff
@@ -85,6 +87,45 @@ func getAutoScalingGroupsByName(service autoscalingiface.AutoScalingAPI, autoSca
 	return result, nil
 }
 
+func getOwnedAutoScalingGroups(service autoscalingiface.AutoScalingAPI) (map[string]*autoScalingGroupDetails, error) {
+	params := &autoscaling.DescribeAutoScalingGroupsInput{}
+	resp, err := service.DescribeAutoScalingGroups(params)
```
mikkeloscar: I think we need to use: https://docs.aws.amazon.com/sdk-for-go/api/service/autoscaling/#AutoScaling.DescribeAutoScalingGroupsPages instead because `DescribeAutoscalingGroups` will only list a limited number of autoscaling groups (The limit is `100`, but still we should handle this edge case).

aws/asg.go
```diff
@@ -85,6 +87,45 @@ func getAutoScalingGroupsByName(service autoscalingiface.AutoScalingAPI, autoSca
...
+	if err != nil {
+		return nil, err
+	}
+
+	result := make(map[string]*autoScalingGroupDetails)
+	for _, g := range resp.AutoScalingGroups {
+		name := aws.StringValue(g.AutoScalingGroupName)
+
+		isOwn := false
```
mikkeloscar: nit: I would call this `isOwned` :)

aws/asg.go
```diff
@@ -85,6 +87,45 @@ func getAutoScalingGroupsByName(service autoscalingiface.AutoScalingAPI, autoSca
...
+			value := aws.StringValue(td.Value)
+			tags[key] = value
+
+			if strings.HasPrefix(key, clusterIDTagPrefix) && value == resourceLifecycleOwned {
+				isOwn = true
+			}
+		}
+
+		if !isOwn {
+			continue
```
mikkeloscar: nit: I think the code will be a bit simpler to follow if you do:

```go
if isOwn {
  result[name] = ...
}
```

aws/adapter.go
gchudnov: got it, my fault. will fix.

aws/asg.go
gchudnov: aha, will fix

aws/adapter.go
```diff
@@ -636,39 +636,22 @@ func (a *Adapter) UpdateAutoScalingGroupsAndInstances() error {
...
-		if err != nil {
-			log.Errorf("failed fetching Auto Scaling Groups details: %v", err)
-		} else {
-			for name, asg := range fetchedAutoScalingGroups {
-				newAutoScalingGroups[name] = asg
-			}
+	newAutoScalingGroups := make(map[string]*autoScalingGroupDetails)
+	fetchedAutoScalingGroups, err := getOwnedAutoScalingGroups(a.autoscaling)
+	if err != nil {
+		log.Errorf("failed fetching Auto Scaling Groups details: %v", err)
```
mikkeloscar: We should return this as an error instead of logging it here to be safe. If we just log it then we will set `a.autoScalingGroups` to an empty map below which would be wrong. Before logging kind a worked because it would just set the map to the same value as it already was in case of error on this call.

mikkeloscar:
Hi, thanks for following up on the comments. I added one more now but other than that it looks good to me! Apologies for the slow review!

aws/adapter.go
gchudnov: yes, you're right. changing..

aws/adapter.go
```diff
@@ -636,39 +636,22 @@ func (a *Adapter) UpdateAutoScalingGroupsAndInstances() error {
...
-			log.Errorf("failed fetching Auto Scaling Groups details: %v", err)
-		} else {
-			for name, asg := range fetchedAutoScalingGroups {
-				newAutoScalingGroups[name] = asg
-			}
+	newAutoScalingGroups := make(map[string]*autoScalingGroupDetails)
+	fetchedAutoScalingGroups, err := getOwnedAutoScalingGroups(a.autoscaling)
+	if err != nil {
+		return err
+	} else {
```
golangcibot: `if` block ends with a `return` statement, so drop this `else` and outdent its block (from `golint`)

mikkeloscar:
:+1:

gchudnov:
@mikkeloscar 
thank you for +1. What is the process in general here? Is it required by someone one to review the PR?

szuecs:
@gchudnov the process is 3 eyes from the team maintaining this project. I will review it today and thanks for the PR.

szuecs:
:+1:

PR #249, opened 2019-05-02
Drop Docker stage build to speed up build

mikkeloscar:
Drops the docker stage build as it's much slower because you need to pull the `golang` image on the CI and re-pull all dependencies as they are not cached by CDP.

You also need to re-pull dependencies every time when running locally.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,6 @@
-# builder image
-FROM golang as builder
-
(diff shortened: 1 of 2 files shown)


mikkeloscar:
:+1:

hjacobs:
:+1:

Issue #248, opened 2019-05-02
v0.8.4 can not attach member to ASG


szuecs:
I think we merged a bug in https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/245

```
time="2019-05-02T11:13:44Z" level=error msg="UpdateTargetGroupsAndAutoScalingGroups() failed to attach target groups to ASG: ValidationError: 1 validation error detected: Value '[]' at 'targetGroupARNs' failed to satisfy constraint: Member must satisfy constraint: [Member must have length less than or equal to 511, Member must have length greater than or equal to 1, Member must satisfy regular expression pattern: [\\u0020-\\uD7FF\\uE000-\\uFFFD\\uD800\\uDC00-\\uDBFF\\uDFFF\\r\\n\\t]*]\n\tstatus code: 400, request id: 598e9cd1-6ccb-11e9-81e2-c504391d57ef"
time="2019-05-02T11:13:44Z" level=error msg="UpdateTargetGroupsAndAutoScalingGroups() failed to attach target groups to ASG: ValidationError: 1 validation error detected: Value '[]' at 'targetGroupARNs' failed to satisfy constraint: Member must satisfy constraint: [Member must have length less than or equal to 511, Member must have length greater than or equal to 1, Member must satisfy regular expression pattern: [\\u0020-\\uD7FF\\uE000-\\uFFFD\\uD800\\uDC00-\\uDBFF\\uDFFF\\r\\n\\t]*]\n\tstatus code: 400, request id: 5998aedf-6ccb-11e9-b9ea-19a93c0c3424"
time="2019-05-02T11:13:44Z" level=error msg="UpdateTargetGroupsAndAutoScalingGroups() failed to attach target groups to ASG: ValidationError: 1 validation error detected: Value '[]' at 'targetGroupARNs' failed to satisfy constraint: Member must satisfy constraint: [Member must have length less than or equal to 511, Member must have length greater than or equal to 1, Member must satisfy regular expression pattern: [\\u0020-\\uD7FF\\uE000-\\uFFFD\\uD800\\uDC00-\\uDBFF\\uDFFF\\r\\n\\t]*]\n\tstatus code: 400, request id: 59a1d63e-6ccb-11e9-b3f0-6bd4f0e88fa5"
time="2019-05-02T11:13:44Z" level=error msg="Failed to update ingress: invalid ingress update parameters"
[...]

PR #250, opened 2019-05-02
Correctly lookup asg clusterID ownership

mikkeloscar:
Fixes a bug introduced in #245 (https://github.com/zalando-incubator/kube-ingress-aws-controller/issues/248) where only the clusterID prefix was checked, resulting in the controller collecting *ALL* ASGs in the account including those not part of the cluster.

Additionally it checks that the stack is ready before trying to update ingress or target groups. Before this would lead to errors until the stack was fully created in AWS.


szuecs:
all tests are broken...

mikkeloscar:
@szuecs fixed

szuecs:
:+1:

mikkeloscar:
:+1:

PR #265, opened 2019-07-18
Don't add others ingresses to already owned/non-shared stack

mikkeloscar:
It seems like we have a bug where shared ingresses can be added to an already non-shared stack owned by another ingress.

Need to add tests to validate that this is indeed a bug.


szuecs:
Added test and validated that it caught the bug.

szuecs:
:+1:

muaazsaleem:
👍

mikkeloscar:
:+1:

PR #268, opened 2019-08-16
Log when security group cannot be found

ruudk:
This would have saved me a couple of hours today ;)

### DIFF
--- a/aws/ec2.go
+++ b/aws/ec2.go
@@ -314,7 +314,7 @@ func findSecurityGroupWithClusterID(svc ec2iface.EC2API, clusterID string, contr
 	}
 
 	if len(resp.SecurityGroups) < 1 {
-		return nil, ErrMissingSecurityGroup
+		return nil, fmt.Errorf("could not find security group that matches: %s", params.Filters)
 	}
 
 	sg := resp.SecurityGroups[0]
--- a/aws/adapter.go
+++ b/aws/adapter.go
@@ -104,8 +104,6 @@ const (
 )
 
 var (
-	// ErrMissingSecurityGroup is used to signal that the required security group couldn't be found.
-	ErrMissingSecurityGroup = errors.New("required security group was not found")
 	// ErrLoadBalancerStackNotFound is used to signal that a given load balancer CF stack was not found.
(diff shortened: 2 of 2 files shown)


szuecs:
:+1:

szuecs:
@ruudk thanks

aws/ec2.go
```diff
@@ -314,6 +314,8 @@ func findSecurityGroupWithClusterID(svc ec2iface.EC2API, clusterID string, contr
 	}
 
 	if len(resp.SecurityGroups) < 1 {
+		log.Errorf("Cannot find security group that matches: %v\n", params)
+
 		return nil, ErrMissingSecurityGroup
```
mikkeloscar: This error is already logged as `Fatal`. I suggest we just enhance this error to what you put above if it's not clear enough as is. otherwise we log related errors twice.

ruudk: I was thinking the same, but how to do that? Since the error is a typed var. Could I just return a new error with this formatted in?

mikkeloscar: > Could I just return a new error with this formatted in?

Yes, we only use the var in this place so we can just drop it and return a formatted error.

ruudk: Fixed, the error will now look like this:
```
could not find security group that matches: [{
  Name: "tag-key",
  Values: ["kubernetes.io/cluster/foo"]
} {
  Name: "tag-value",
  Values: ["owned"]
} {
  Name: "tag-key",
  Values: ["kubernetes:application"]
} {
  Name: "tag-value",
  Values: ["kube-ingress-aws-controller"]
}]
```

mikkeloscar:
:+1:

szuecs:
Can you sign-off the commit please?
git commit —ammend—sign-off as far as I remember without checking.

ruudk:
I did that , right? What’s the purpose of that btw? This is the first project that wants that.

ruudk:
@szuecs Did it again. Let's ship it :) ?

mikkeloscar:
:+1:

szuecs:
:+1:

Issue #271, opened 2019-09-11
Support for Cloudwatch Alarm configuration


martinohmann:
We'd like to automatically set up Cloudwatch Alarms for certain ALB metrics (e.g. `RejectedConnectionCount`) to be able to get better visibility of problems that might occur between the ALBs and the actual ingress controller (nginx-ingress-controller in our case).

I was thinking about making Cloudwatch Alarms configurable in an optional ConfigMap which would then be rendered into the Cloudformation template for the ALBs.

Is this a feature that you would find useful? I'd be willing to work on it and provide you with a PR.

szuecs:
@martinohmann can you show how the data could look like that you imagine to configure Cloudwatch Alarms for an ingress?

Right now I would recommend to try to use ingress annotations, because the state is only in one place, but maybe it makes sense in this case. :)

martinohmann:
Sadly, ingress annotations are not flexible enough for this feature because each Cloudwatch Alarm has quite a few fields that can be/need to be configured.

Here is the documentation for the CW Alarm fields that can be set via Cloudformation: https://docs.aws.amazon.com/en_us/AWSCloudFormation/latest/UserGuide/aws-properties-cw-alarm.html#aws-properties-cw-alarm-syntax.yaml

There are basically only two fields that can be left out, because they can be inferred while generating the cloudformation template: `Namespace` is always `AWS/ApplicationELB` and `Dimension` would only contain one key-value pair which is `LoadBalancerName: <generated-alb-name>` and which would be filled while rendering the CF template.

A ConfigMap for that could look like this:
```
---
kind: ConfigMap
apiVersion: v1
metadata:
  name: kube-ingress-aws-controller-alarms
  namespace: kube-system
data:
  alarms: |
    - ActionsEnabled: true
      AlarmActions: 
      - arn:aws:sns:region:account-id:sns-topic-name
      AlarmDescription: "Some description of the alarm"
      AlarmName: "Some alarm"
      ComparisonOperator: GreaterThanOrEqualToThreshold
      DatapointsToAlarm: 2
      EvaluateLowSampleCountPercentile: evaluate
      EvaluationPeriods: 1
      ExtendedStatistic: String
      InsufficientDataActions: 
      - arn:aws:sns:region:account-id:sns-topic-name
      MetricName: RejectedConnectionCount
      OKActions: 
      - arn:aws:sns:region:account-id:sns-topic-name
      Period: 600
      Statistic: Maximum
[...]

PR #272, opened 2019-09-13
Support for CloudWatch Alarm configuration

martinohmann:
This PR adds support for CloudWatch Alarm configuration (#271) for each load
balancer created by the controller. This is an opt-in feature. To use
it, the `--cloudwatch-alarms-config-map` CLI flag has to be set and
should be pointed to a ConfigMap that contains the alarm configuration.
If the ConfigMap does not exist or contains malformed data, the
controller will just ignore it. Check out `cloudwatch.md` for a full
documentation of the feature.

### DIFF
--- a/aws/cf.go
+++ b/aws/cf.go
@@ -26,10 +26,10 @@ type Stack struct {
 	SecurityGroup     string
 	SSLPolicy         string
 	IpAddressType     string
-	TargetGroupARN    string
-	CertificateARNs   map[string]time.Time
 	OwnerIngress      string
 	CWAlarmConfigHash string
+	TargetGroupARN    string
+	CertificateARNs   map[string]time.Time
 	tags              map[string]string
 }
 
--- a/aws/cf.go
+++ b/aws/cf.go
@@ -3,7 +3,6 @@ package aws
 import (
 	"fmt"
 	"strings"
-
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -15,21 +14,23 @@ const (
 	certificateARNTagLegacy = "ingress:certificate-arn"
 	certificateARNTagPrefix = "ingress:certificate-arn/"
 	ingressOwnerTag         = "ingress:owner"
+	cwAlarmConfigHashTag    = "cloudwatch:alarm-config-hash"
 )
 
 // Stack is a simple wrapper around a CloudFormation Stack.
 type Stack struct {
-	Name            string
-	status          string
-	DNSName         string
-	Scheme          string
-	SecurityGroup   string
-	SSLPolicy       string
-	IpAddressType   string
-	TargetGroupARN  string
-	CertificateARNs map[string]time.Time
-	OwnerIngress    string
-	tags            map[string]string
+	Name              string
+	status            string
+	DNSName           string
+	Scheme            string
+	SecurityGroup     string
+	SSLPolicy         string
+	IpAddressType     string
+	TargetGroupARN    string
+	CertificateARNs   map[string]time.Time
+	OwnerIngress      string
+	CWAlarmConfigHash string
+	tags              map[string]string
 }
 
 // IsComplete returns true if the stack status is a complete state.
@@ -132,6 +133,7 @@ type stackSpec struct {
 	albLogsS3Bucket              string
 	albLogsS3Prefix              string
 	wafWebAclId                  string
+	cwAlarms                     CloudWatchAlarmList
 }
 
 type healthCheck struct {
@@ -141,7 +143,7 @@ type healthCheck struct {
 }
 
 func createStack(svc cloudformationiface.CloudFormationAPI, spec *stackSpec) (string, error) {
-	template, err := generateTemplate(spec.certificateARNs, spec.idleConnectionTimeoutSeconds, spec.albLogsS3Bucket, spec.albLogsS3Prefix, spec.wafWebAclId)
+	template, err := generateTemplate(spec)
 	if err != nil {
 		return "", err
 	}
@@ -183,6 +185,10 @@ func createStack(svc cloudformationiface.CloudFormationAPI, spec *stackSpec) (st
 		params.Tags = append(params.Tags, cfTag(ingressOwnerTag, spec.ownerIngress))
 	}
 
+	if len(spec.cwAlarms) > 0 {
+		params.Tags = append(params.Tags, cfTag(cwAlarmConfigHashTag, spec.cwAlarms.Hash()))
+	}
+
 	resp, err := svc.CreateStack(params)
 	if err != nil {
 		return spec.name, err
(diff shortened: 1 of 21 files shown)


szuecs:
Thanks @martinohmann for the PR. Have to carefully review, which I hopefully do next week.

martinohmann:
If you want to try out the PR on a real cluster, you can use the following example ConfigMap which creates a CloudWatch Alarm that triggers when the TargetGroup of the LoadBalancer has no healthy hosts:

```yaml
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: kube-ingress-aws-controller-cw
  namespace: kube-system
data:
  alarms: |
    - ActionsEnabled: false
      AlarmDescription: "Triggers once the TargetGroup of the LoadBalancer has no healthy hosts."
      AlarmName: "HealthyHosts"
      ComparisonOperator: LessThanThreshold
      Dimensions:
      - Name: LoadBalancer
      - Name: TargetGroup
      EvaluationPeriods: 1
      MetricName: HealthyHostCount
      Period: 60
      Statistic: Average
      Threshold: 1
      TreatMissingData: notBreaching
      Unit: Count
```

Be sure to also grant the controller the necessary IAM permissions as written in the docs and deploy the adjusted ClusterRole which allows reading ConfigMaps. The ConfigMap just sets up the Alarm but does not set up any alarm actions (e.g. push to SNS) as this would require more permission setup and is out of scope.

aws/cf.go
```diff
@@ -15,21 +14,23 @@ const (
...
+	status            string
+	DNSName           string
+	Scheme            string
+	SecurityGroup     string
+	SSLPolicy         string
+	IpAddressType     string
+	TargetGroupARN    string
+	CertificateARNs   map[string]time.Time
+	OwnerIngress      string
+	CWAlarmConfigHash string
```
szuecs: Can you move the 2 string members OwnerIngress and CWAlarmConfigHash just before TargetGroupARN, please?
This will pack the struct better.

martinohmann: Sure thing, fixed!

aws/cf.go
```diff
@@ -342,17 +352,18 @@ func mapToManagedStack(stack *cloudformation.Stack) *Stack {
...
+		TargetGroupARN:    outputs.targetGroupARN(),
+		Scheme:            parameters[parameterLoadBalancerSchemeParameter],
+		SecurityGroup:     parameters[parameterLoadBalancerSecurityGroupParameter],
+		SSLPolicy:         parameters[parameterListenerSslPolicyParameter],
+		IpAddressType:     parameters[parameterIpAddressTypeParameter],
+		CertificateARNs:   certificateARNs,
+		tags:              tags,
+		OwnerIngress:      ownerIngress,
+		status:            aws.StringValue(stack.StackStatus),
+		CWAlarmConfigHash: tags[cwAlarmConfigHashTag],
```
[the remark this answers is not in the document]
martinohmann: It is possible that we do not have the `cwAlarmConfigHashTag`, which is the case when no CloudWatch Alarms are configured. In this case `CWAlarmConfigHash` is empty, which is a perfectly valid value as it represents the hash of an empty CloudWatch Alarm configuration.

aws/cloudwatch.go
```diff
@@ -0,0 +1,102 @@
+package aws
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
+
+	"github.com/ghodss/yaml"
```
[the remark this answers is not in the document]
martinohmann: The `go-cloudformation` package only has the `json` struct tag set for the field names on its types, but not the `yaml` one. `gopkg.in/yaml.v2` does not support the `json` struct tag, which causes the alarm configuration to unmarshal into empty `cloudformation.CloudWatchAlarm` structs. `github.com/ghodss/yaml` addresses this issue and is generally the safer bet.

aws/cloudwatch.go
```diff
@@ -0,0 +1,102 @@
...
+func (c CloudWatchAlarmList) Hash() string {
+	if len(c) == 0 {
+		return ""
+	}
+
+	hash := sha256.New()
+
+	buf, _ := json.Marshal(c)
+
+	hash.Write(buf)
```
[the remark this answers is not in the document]
martinohmann: You are right about at least logging the `json.Marshal` error, i fixed that. However, for `hash.Write` I have to disagree. It just returns an `error` to fulfill the `io.Writer` interface, but this error is guaranteed to be always `nil`, as you can see in the source of the `crypto/sha256` package: https://github.com/golang/go/blob/master/src/crypto/sha256/sha256.go#L184-L205

So I think it is pointless to check for an error that will never happen :).

You're already doing the same thing in `hashARNs`.

aws/cloudwatch.go
```diff
@@ -0,0 +1,102 @@
...
+
+// normalizeCloudWatchAlarmDimensions replaces the values of the dimensions
+// LoadBalancer and TargetGroup (if set) with the generated names of these
+// resources. If alarmDimensions is nil or empty, it will return a sane default
+// that includes the dimension LoadBalancer with its generated name.
+func normalizeCloudWatchAlarmDimensions(alarmDimensions *cloudformation.CloudWatchAlarmDimensionList) *cloudformation.CloudWatchAlarmDimensionList {
+	if alarmDimensions == nil || len(*alarmDimensions) == 0 {
+		// For convenience, include LoadBalancer in the dimensions if the user
+		// provided nothing in the configuration. This is the dimension we want
+		// to use most of the time.
```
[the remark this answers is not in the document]
martinohmann: I removed the whole doc-block, as the first part is already mentioned in the doc-block of the function.

szuecs:
:+1:

szuecs:
@mikkeloscar please review next week, thanks

szuecs:
@martinohmann great PR, thanks!

worker.go
```diff
@@ -242,14 +244,26 @@ func doWork(certsProvider certs.CertificatesProvider, certsPerALB int, certTTL t
...
 	}
 
+	cwAlarms, err := getCloudWatchAlarms(kubeAdapter, cwAlarmConfigMapLocation)
+	if err != nil {
+		// We will just log potential errors while loading the alarm
+		// configuration. A non-existent config map or missing service account
+		// permissions should not cause the ingress controller to stop working.
+		// Instead, we just treat the configuration as non-existent and move
+		// on.
+		log.Warnf("ignoring cloudwatch alarm configuration due to error: %v", err)
```
mikkeloscar: I don't agree with this. We should be explicit about the errors and not hide them as warnings IMO. I would rather we only try to lookup a configmap if the use specifies it via the flag, and if not we don't attempt the get config call at all.

martinohmann: I get your point. However, if you check the `getCloudWatchAlarms` func you can see that it only performs config lookup if the config map location is non-nil (which is only the case when the user provided it explicitly). How should potential errors be handled here? We could just log the error with level Error and move on (as in the current implementaion) or return it. The latter case would cause the controller to stop working.

martinohmann: Got it! Will change that tomorrow.

martinohmann: Pushed a change just now.

mikkeloscar:
:+1:

szuecs:
:+1:

martinohmann:
@szuecs @mikkeloscar thanks for merging! Is it possible to get a new tagged release for this?

szuecs:
@martinohmann https://github.com/zalando-incubator/kube-ingress-aws-controller/releases/tag/v0.8.13, thanks again for the PR!

martinohmann:
Thanks alot!

PR #275, opened 2019-10-01
Restore CUSTOM_FILTERS function which was broken in 0.8.4.

### LINKED ISSUE #266: Regression in CUSTOM_FILTERS >= v0.8.4
We have been using the CUSTOM_FILTERS mechanism to isolate a certain set of ASGs as valid for target groups.

Versions >= v0.8.4 no longer honour the CUSTOM_FILTERS specified and instead include all non-master ASGs, resulting in extra, unhealthy hosts in the target groups list.

I haven't had a chance to do deep analysis yet, but the new getOwnedAutoScalingGroups func seems to be at the heart of the issue, and does not consult the filter

jhohertz:
This fixes #266, which describes the issue being addressed here. In short, CUSTOM_FILTERS is only applied to DescribeInstances, so ensure we limit the ASGs considered to those seen in the described instances.

It was originally considered to try to apply tag filtering to the ASG lookups as well, but in the end, simply restored the prior logic with a minimal patch.

Signed-off-by: Joe Hohertz <joe@viafoura.com>

### DIFF
--- a/aws/asg.go
+++ b/aws/asg.go
@@ -85,7 +85,7 @@ func getAutoScalingGroupsByName(service autoscalingiface.AutoScalingAPI, autoSca
 	return result, nil
 }
 
-func getOwnedAutoScalingGroups(service autoscalingiface.AutoScalingAPI, clusterID string) (map[string]*autoScalingGroupDetails, error) {
+func getOwnedAutoScalingGroups(service autoscalingiface.AutoScalingAPI, clusterID string, asgNameMap map[string]bool) (map[string]*autoScalingGroupDetails, error) {
 	params := &autoscaling.DescribeAutoScalingGroupsInput{}
 
 	clusterIDTag := clusterIDTagPrefix + clusterID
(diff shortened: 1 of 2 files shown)


aws/asg.go
```diff
@@ -103,7 +103,7 @@ func getOwnedAutoScalingGroups(service autoscalingiface.AutoScalingAPI, clusterI
 					value := aws.StringValue(td.Value)
 					tags[key] = value
 
-					if key == clusterIDTag && value == resourceLifecycleOwned {
+					if key == clusterIDTag && value == resourceLifecycleOwned && asgNameMap[name] {
```
mikkeloscar: This unfortunately breaks what the other PR introducing the regression fixed. Which is find ASGs which are managed but currently have 0 running instances.

It seems to me these two features are somewhat conflicting. At least in the way we are trying to fix them right now.

I think the only right way is to look at ASG tags and see if they match the filter.

jhohertz: Hmmmm. I was a little worried about that.... I had taken a couple of runs at adding tag filters to ASG, but I really didn't like the shape of what was coming out (largely duplicating functions used for the instance filtering, but which I couldn't re-use due to slightly different types in the AWS APIs), and came back to this.

I'll keep pinning to 0.8.3 and see if I have time to try again soon.

jhohertz:
Closing in favour of #277

Issue #199, opened 2018-08-02
Add support for HTTP -> HTTPS redirect.


mikkeloscar:
Now supported directly on the ALB: https://aws.amazon.com/about-aws/whats-new/2018/07/elastic-load-balancing-announces-support-for-redirects-and-fixed-responses-for-application-load-balancer/

hjacobs:
We don't really need that for Zalando as we are using Skipper, but it would still be very nice to have :smile:

so0k:
Skipper has been great so far, but a new requirement to route traffic through a Firewall DaemonSet on the edge nodes no longer allows us to run Skipper with hostPort exposed. 

As for the design of this feature, I see 2 options:

1. Just use custom CF Template or add flag for controller
1. Implement `ingress.kubernetes.io/force-ssl-redirect: "true"` Ingress annotation

### Option 1: Custom template or controller flag

We could just use a custom CF template which sets port 80 default action to redirect https

To make it more flexible, we could just add a `-force-ssl-redirect` controller flag and update the built-in CF template to take a boolean template variable. 

### Option 2: Implement ssl-redirect Ingress annotation

To avoid https://github.com/heptio/contour/issues/250 the `buildManagedModel` would need to create a different ALB for Ingress objects without the annotation. The CloudFormation Template would need an additional parameter

For my use case, Option 2 wouldn't work because I do all TLS termination with ACM+ALB and https://github.com/heptio/contour/issues/715 -> Contour and kube-ingress-aws-controller both act on the same ingress objects (they share ingress class) so Contour updates Envoy HTTP listeners to redirect to HTTPS. Envoy gets HTTP from ALB and goes into a redirect loop. Ideally Envoy recognises [X-Forwarded-Proto](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto) to avoid the redirect loop.

Option 1 is also much easier to implement, would a PR to add a controller flag be useful?

szuecs:
@so0k I am not against any change enabling your use case, but I am also interested in the problem in detail you had with the skipper deployment  in detail. Would you mind creating an issue in skipper repository, too?
I really would like to understand if there’s a limitation or we need more documentation how to support your use case.

To add my opinion on your options, I think both options could be useful and can support different use cases. 
[...]

PR #276, opened 2019-10-02
implement an http-to-https redirect option

jhohertz:
Implements the feature described in #199

Simply adds a controller flag, that if set, replaces the forward rule on the HTTPListener with a redirection rule to HTTPS, preserving the rest of the URL.

Note for reviewer: This did require me to update go-cloudformation (and it's dependencies), as `cloudformation.ElasticLoadBalancingV2ListenerRedirectConfig` was not present in the pinned version.

### DIFF
--- a/aws/adapter.go
+++ b/aws/adapter.go
@@ -293,7 +293,7 @@ func (a *Adapter) WithWafWebAclId(wafWebAclId string) *Adapter {
 }
 
 // WithHttpRedirectToHttps returns the receiver adapter after changing the flag to effect HTTP->HTTPS redirection
-func (a *Adapter) WithHttpRedirectToHttps (httpRedirectToHttps bool) *Adapter {
+func (a *Adapter) WithHttpRedirectToHttps(httpRedirectToHttps bool) *Adapter {
 	a.httpRedirectToHttps = httpRedirectToHttps
 	return a
 }
--- a/aws/adapter.go
+++ b/aws/adapter.go
@@ -59,6 +59,7 @@ type Adapter struct {
 	albLogsS3Bucket            string
 	albLogsS3Prefix            string
 	wafWebAclId                string
+	httpRedirectToHttps        bool
 }
 
 type manifest struct {
@@ -291,6 +292,12 @@ func (a *Adapter) WithWafWebAclId(wafWebAclId string) *Adapter {
 	return a
 }
 
+// WithHttpRedirectToHttps returns the receiver adapter after changing the flag to effect HTTP->HTTPS redirection
+func (a *Adapter) WithHttpRedirectToHttps (httpRedirectToHttps bool) *Adapter {
+	a.httpRedirectToHttps = httpRedirectToHttps
+	return a
+}
+
 // ClusterID returns the ClusterID tag that all resources from the same Kubernetes cluster share.
 // It's taken from the current ec2 instance.
 func (a *Adapter) ClusterID() string {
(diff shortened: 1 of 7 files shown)


jhohertz:
Fixed the missed signoff

aws/adapter.go
```diff
@@ -291,6 +292,12 @@ func (a *Adapter) WithWafWebAclId(wafWebAclId string) *Adapter {
 	return a
 }
 
+// WithHttpRedirectToHttps returns the receiver adapter after changing the flag to effect HTTP->HTTPS redirection
+func (a *Adapter) WithHttpRedirectToHttps (httpRedirectToHttps bool) *Adapter {
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
func (a *Adapter) WithHttpRedirectToHttps(httpRedirectToHttps bool) *Adapter {
```

aws/cf_template.go
```diff
@@ -74,17 +74,38 @@ func generateTemplate(spec *stackSpec) (string, error) {
...
-	template.AddResource("HTTPListener", &cloudformation.ElasticLoadBalancingV2Listener{
-		DefaultActions: &cloudformation.ElasticLoadBalancingV2ListenerActionList{
-			{
-				Type:           cloudformation.String("forward"),
-				TargetGroupArn: cloudformation.Ref("TG").String(),
+	if spec.httpRedirectToHttps {
+		template.AddResource("HTTPListener", &cloudformation.ElasticLoadBalancingV2Listener{
+			DefaultActions: &cloudformation.ElasticLoadBalancingV2ListenerActionList{
+				{
+					Type:           cloudformation.String("redirect"),
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
					Type: cloudformation.String("redirect"),
```

jhohertz:
Fix the fact the bot doesn't apply DCO if applying suggestions

mikkeloscar:
@jhohertz thanks for this, makes a lot of sense to support it here.

The PR looks good to me, but could you mention this feature in the README? :)

mikkeloscar:
Maybe also a reference to the docs in the code would be helpful to understand how these magic variables work: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-elasticloadbalancingv2-listener-redirectconfig.html I had to search for this to understand it.

jhohertz:
Force push was due to forgetting to `--sign` the commit. :man_facepalming:

mikkeloscar:
Thanks! LGTM.

@szuecs could we also use this in our setup?

szuecs:
@mikkeloscar from my understanding, yes we can replace skipper's redirect with this one.
I think this should be also the default in the install guide, which we can change later.

@jhohertz thanks for your great work!

szuecs:
:+1:

szuecs:
Unclear to me from the docs is if https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/276/files#diff-728b77723ddf60461dfa4f0107b891c8R85 with empty string works as expected for example. We should test this in a cluster.

mikkeloscar:
:+1:

jhohertz:
Thanks for merging this in! :)

tomekit:
Is there a possibility to add any annotation on the Ingress level in order to enable HTTP->HTTPS redirection on the ALB level?

mikkeloscar:
This is not available. We treat this setting as a global setting for the whole ingress stack. However you're welcome to open an issue to discuss the use case and possibility of adding this as a feature.

szuecs:
I agree with @mikkeloscar .
@tomekit you can work around it with using non shared ALBs, if you like. It costs more, but it would be based on ingress.

PR #277, opened 2019-10-02
Rework fix for CUSTOM_FILTERS

jhohertz:
This PR is to address issue #266 and obsoletes PR #275.

It should not cause a regression to the patches which caused this regression. :)

Signed-off-by: Joe Hohertz <joe@viafoura.com>

### DIFF
--- a/aws/adapter.go
+++ b/aws/adapter.go
@@ -6,8 +6,6 @@ import (
 	"strings"
 	"time"
 
-	"os"
-
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/client"
 	"github.com/aws/aws-sdk-go/aws/ec2metadata"
@@ -60,6 +58,7 @@ type Adapter struct {
 	albLogsS3Prefix            string
 	wafWebAclId                string
 	httpRedirectToHttps        bool
+	customFilter               string
 }
 
 type manifest struct {
@@ -98,11 +97,10 @@ const (
 	// DefaultAlbS3LogsPrefix is a blank string, and optionally set if desired
 	DefaultAlbS3LogsPrefix = ""
 	DefaultWafWebAclId     = ""
+	DefaultCustomFilter    = ""
 
 	ipAddressTypeDualstack = "dualstack"
 	nameTag                = "Name"
-
-	customTagFilterEnvVarName = "CUSTOM_FILTERS"
 )
 
 var (
@@ -175,6 +173,7 @@ func NewAdapter(newControllerID string) (adapter *Adapter, err error) {
 		albLogsS3Bucket:     DefaultAlbS3LogsBucket,
 		albLogsS3Prefix:     DefaultAlbS3LogsPrefix,
 		wafWebAclId:         DefaultWafWebAclId,
+		customFilter:        DefaultCustomFilter,
 	}
 
 	adapter.manifest, err = buildManifest(adapter)
@@ -297,6 +296,15 @@ func (a *Adapter) WithWafWebAclId(wafWebAclId string) *Adapter {
 // WithHttpRedirectToHttps returns the receiver adapter after changing the flag to effect HTTP->HTTPS redirection
 func (a *Adapter) WithHttpRedirectToHttps(httpRedirectToHttps bool) *Adapter {
 	a.httpRedirectToHttps = httpRedirectToHttps
+	return a
+}
+
+// WithCustomFilter returns the receiver adapter after setting a custom filter expression
+func (a *Adapter) WithCustomFilter(customFilter string) *Adapter {
+	a.customFilter = customFilter
+	// also rebuild related manifest items
+	a.manifest.filters = a.parseFilters(a.ClusterID())
+	a.manifest.asgFilters = a.parseAutoscaleFilterTags(a.ClusterID())
 	return a
 }
 
@@ -576,8 +584,8 @@ func buildManifest(awsAdapter *Adapter) (*manifest, error) {
 		securityGroup: securityGroupDetails,
 		instance:      instanceDetails,
 		subnets:       subnets,
-		filters:       parseFilters(clusterID),
-		asgFilters:    parseAutoscaleFilterTags(clusterID),
+		filters:       awsAdapter.parseFilters(clusterID),
+		asgFilters:    awsAdapter.parseAutoscaleFilterTags(clusterID),
 	}, nil
 }
 
@@ -699,14 +707,14 @@ func (a *Adapter) UpdateAutoScalingGroupsAndInstances() error {
 // later on each cycle. Filter is based on value of customTagFilterEnvVarName environment
 // veriable. If it is undefined or could not be parsed, default filter is returned which
 // filters on kubernetesClusterTag tag value and kubernetesNodeRoleTag existance.
-func parseFilters(clusterId string) []*ec2.Filter {
-	if filter, ok := os.LookupEnv(customTagFilterEnvVarName); ok {
-		terms := strings.Fields(filter)
+func (a *Adapter) parseFilters(clusterId string) []*ec2.Filter {
+	if a.customFilter != "" {
+		terms := strings.Fields(a.customFilter)
 		filters := make([]*ec2.Filter, len(terms))
 		for i, term := range terms {
 			parts := strings.Split(term, "=")
 			if len(parts) != 2 {
-				log.Errorf("failed parsing %s, falling back to default", customTagFilterEnvVarName)
+				log.Errorf("failed parsing %s, falling back to default", a.customFilter)
 				return generateDefaultFilters(clusterId)
 			}
 			filters[i] = &ec2.Filter{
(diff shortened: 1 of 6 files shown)


aws/asg.go
```diff
@@ -3,6 +3,9 @@ package aws
 import (
 	"context"
 	"fmt"
+	"os"
+	"strings"
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	"strings"

```

jhohertz: Amended commit with this change

jhohertz:
I :heart: playing please the linter... No idea what it was on about, but did a gofmt on the file, and things seem happy now.

szuecs:
@jhohertz can you please create table driven unit tests for these 2 functions?

jhohertz:
Will do.

aws/asg.go
```diff
@@ -85,30 +89,86 @@ func getAutoScalingGroupsByName(service autoscalingiface.AutoScalingAPI, autoSca
 	return result, nil
 }
 
+// We look to the same string used for instance filtering, however, we are much more limited in what can be done for
+// ASGs. As such, we instead build a map of tags to look for as we iterate over all ASGs in getOwnedAutoScalingGroups
+func parseFilterTags(clusterId string) map[string][]string {
+	if filter, ok := os.LookupEnv(customTagFilterEnvVarName); ok {
```
mikkeloscar: I would prefer if we could handle `os.LookupEnv` and value parsing in the beginning of the program i.e. close to `main` func where we also parse the flags. This allows us to keep such functions as these pure and simpler to test.

Given that we already have something similar here: https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/bd660aa5cbeafd030572afc083c8aba21c73d81a/aws/adapter.go#L686 I'm fine not fixing it in this PR. but I would like to fix it afterwards at least.

aws/asg_test.go
```diff
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"reflect"
 	"testing"
+	"os"
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion

```

aws/asg.go
jhohertz: I could create a new field in the adapter manifest, and the parsing to adapter.go, and add a param to getOwnedAutoScalingGroups for the parsed version to work with?

aws/asg_test.go
jhohertz: This continues to baffle me. I clearly need this import, why is it suggesting I remove it?

aws/asg.go
mikkeloscar: I would handle the parsing as early as here like we do for other configuration: https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/bd660aa5cbeafd030572afc083c8aba21c73d81a/controller.go#L67 and then pass the resulting map to the adapter like you suggest.

aws/asg_test.go
szuecs: Just ignore the bot, if it does not makes sense, like here.
Maybe it'S because of this:
```
$ make lint
golangci-lint run ./...
aws/asg_test.go:4: File is not `goimports`-ed (goimports)
	"fmt"
Makefile:23: recipe for target 'lint' failed
make: *** [lint] Error 1
The command "make lint" exited with 2.
```

jhohertz:
Adjusted so the parsing bits are a bit better positioned (as the existing instance filters are handled), and rebased to current master

aws/adapter.go
```diff
@@ -725,3 +727,38 @@ func generateDefaultFilters(clusterId string) []*ec2.Filter {
 		},
 	}
 }
+
+// We look to the same string used for instance filtering, however, we are much more limited in what can be done for
+// ASGs. As such, we instead build a map of tags to look for as we iterate over all ASGs in getOwnedAutoScalingGroups
+func parseAutoscaleFilterTags(clusterId string) map[string][]string {
+       if filter, ok := os.LookupEnv(customTagFilterEnvVarName); ok {
```
golangcibot: File is not `goimports`-ed (from `goimports`)

aws/adapter_test.go
```diff
@@ -767,3 +767,76 @@ func TestFindLBSubnets(tt *testing.T) {
 		})
 	}
 }
+
+func TestParseFilterTagsDefault(t *testing.T) {
+        for _, test := range []struct {
```
golangcibot: File is not `goimports`-ed (from `goimports`)

aws/asg_test.go
```diff
@@ -475,3 +476,58 @@ func TestDetach(t *testing.T) {
 	}
 
 }
+
+func TestTestFilterTags(t *testing.T) {
+	for _, test := range []struct {
+		name          string
```
golangcibot: File is not `goimports`-ed (from `goimports`)

aws/asg.go
```diff
@@ -85,30 +85,51 @@ func getAutoScalingGroupsByName(service autoscalingiface.AutoScalingAPI, autoSca
...
+		}
+	}
+	if len(*filterTags) == len(matches) {
+		return true
+	}
+	return false
+}
 
-	clusterIDTag := clusterIDTagPrefix + clusterID
+func getOwnedAutoScalingGroups(service autoscalingiface.AutoScalingAPI, filterTags *map[string][]string) (map[string]*autoScalingGroupDetails, error) {
```
mikkeloscar: maps are already passed by reference, so no need to accept a pointer here (drop the `*`)

aws/adapter.go
```diff
@@ -67,6 +67,7 @@ type manifest struct {
 	instance      *instanceDetails
 	subnets       []*subnetDetails
 	filters       []*ec2.Filter
+	asgFilters    map[string][]string
```
mikkeloscar: What is the use case for a list of values here? Maybe I'm not 100% understanding how the EC2 filter API works?

jhohertz: So... the ASG filters are... basically useless. You can only do a boolean OR for multiple filters, so we'd have to take a pass over the results to compare tags anyways.... so since we're already iterating over the set whole set of ASGs, I am just doing a comparison of the tags. (Which I munge from the EC2 filter expression)

aws/asg.go
jhohertz: Thanks, adjusted that now

aws/adapter.go
jhohertz: Here's a clipping from cluster-autoscaler where they *do* use a autoscaling.Filter{}, but then have to iterate the result anyways: https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/aws/auto_scaling.go#L130-L145

So there *might* be a small optimisation here, where we also do both, but then we'd have to create yet another filter representation to carry along.

jhohertz:
Rebased

aws/adapter.go
```diff
@@ -725,3 +727,36 @@ func generateDefaultFilters(clusterId string) []*ec2.Filter {
 		},
 	}
 }
+
+// We look to the same string used for instance filtering, however, we are much more limited in what can be done for
+// ASGs. As such, we instead build a map of tags to look for as we iterate over all ASGs in getOwnedAutoScalingGroups
+func parseAutoscaleFilterTags(clusterId string) map[string][]string {
+	if filter, ok := os.LookupEnv(customTagFilterEnvVarName); ok {
```
szuecs: What if we move the ENV setting to the main package and use this function as method (receiver) of `adapter`?
Then you do not need to use set/unset ENV, which might be error prone in concurrent tests.

mikkeloscar: Yes, I agree we should parse the filter from Env in `controller.go` and just pass the resulting value directly to the adapter.

jhohertz: So in the current state, the parsing of the ENV is closely aligned to what is being done for the instance filters. (setup in buildManifest). I had a look at moving the substance of this PR to controller.go and found a lot of dependencies that would have to change as well (such as having access to the clusterID needed for the parser).

What *if*... in controller.go we took the ENV value and passed it verbatim to a new field in the adapter. Anything previously referencing os.Lookupenv would instead be changed to look at that field?

szuecs: exactly!
Store the ENV in controller.go, pass it to the adapter and store it in the adapter.
Then `func parseAutoscaleFilterTags` -> `func (a *adapter) parseAutoscaleFilterTags` and you have access to the value from ENV in `a.customTagFilter`

jhohertz: Okay, sounds good, I'll work something up shortly.

jhohertz: I think the latest revision should be more along what you're thinking?

jhohertz:
Huh... missed that in my testing. Looking into it.

aws/adapter_test.go
```diff
@@ -147,12 +146,20 @@ func TestParseFilters(tt *testing.T) {
 	} {
 		tt.Run(fmt.Sprintf("%v", test.name), func(t *testing.T) {
 			tt.Log(test.name)
+			a := &Adapter{
+				ec2Details:        map[string]*instanceDetails{},
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
				ec2Details: map[string]*instanceDetails{},
```

szuecs:
@jhohertz to advertise your work and show how to use it, there is also documentation missing. Maybe a section in the README.md :)

Code and tests look good to me

jhohertz:
@szuecs : So strictly speaking, this PR is simply to restore the functionality already described in the [how it works section of README.md](https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/master/README.md#how-it-works) which stopped working in v0.8.4.

That said... There is now a small inaccuracy due to the fact that the autoscale filters API does not implement the same way as the EC2 instance tags. (And further because I avoided even using that API).

The new implementation can not support the entire family of filters of the EC2 API. I've kind of faked the `tag` and `tag-key` prefixes to support existing uses of same. However tags like `vpc` and the rest of the EC2 API are just not available. My recommendation is to change the text:

```
where name is one of names that are recognized by the EC2 API (you can find list here)
```

to

```
where name can be a tag: or tag-key: prefix as would be recognised by the EC2 API
```
I did consider also the possibility of constructing a separate "tag cloud" for comparison, from the actual tags, and additional ones that would include `vpc` and anything else that is both available from the ASG description, AND is listed on the EC2 API, and so could be remapped in this manner. Then we could list in the docs any "ec2-like" filters that are supported.

That said... I seem to be the only one that this feature regression seemed to bother, and we aren't using anything besides "tag" and "tag-key" type filters... so.... maybe it's not worth adding the additional complexity here.

Interested in your, and others, thoughts on this before we finalise it.

szuecs:
From my side you can do the minimal change that you suggested.

jhohertz:
Made the change, dropped mention of the 'vpc' tag since it's not supported, and expanded the examples a bit.

szuecs:
:+1:

mikkeloscar:
:+1:

PR #280, opened 2019-10-08
Better handling of invalid SSL Policy

mikkeloscar:
This improves the handling of SSL policy in case the user specifies an invalid value in the annotation.

Before we would just check for valid SSL Policy on Stack creation and fall back to the default policy in case it wasn't valid. This had the problem that new stacks would be created again and again in case the policy was invalid because each newly created stack wouldn't satisfy the ingress with a different (invalid) SSL Policy defined.

Now we fall back to the default already when reading the ingress. Additionally we add a safer validation to the stack creation part where it will fail if the SSL Policy is invalid.

This is missing some form of error reporting for the users, but I think this should be handled in a separate PR e.g. by adding events.

### DIFF
--- a/kubernetes/adapter.go
+++ b/kubernetes/adapter.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	"github.com/aws/aws-sdk-go/service/elbv2"
+	"github.com/zalando-incubator/kube-ingress-aws-controller/aws"
 )
 
 type Adapter struct {
@@ -120,6 +121,11 @@ func (a *Adapter) newIngressFromKube(kubeIngress *ingress) *Ingress {
 		ipAddressType = ipAddressTypeDualstack
 	}
 
+	sslPolicy := kubeIngress.getAnnotationsString(ingressSSLPolicyAnnotation, a.ingressDefaultSSLPolicy)
+	if _, ok := aws.SSLPolicies[sslPolicy]; !ok {
+		sslPolicy = a.ingressDefaultSSLPolicy
+	}
+
 	return &Ingress{
 		CertificateARN: kubeIngress.getAnnotationsString(ingressCertificateARNAnnotation, ""),
 		Namespace:      kubeIngress.Metadata.Namespace,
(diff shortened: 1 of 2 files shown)


kubernetes/adapter.go
```diff
@@ -35,6 +35,16 @@ var (
 	// ErrInvalidCertificates is returned when the CA certificates required to communicate with the
 	// API server are invalid
 	ErrInvalidCertificates = errors.New("invalid CA certificates")
+	// validSSLPolicy is a map of valid ALB SSL Policies
```
szuecs: Can you add the AWS link to the valid certs here?

mikkeloscar: Added

kubernetes/adapter.go
```diff
@@ -120,6 +131,11 @@ func (a *Adapter) newIngressFromKube(kubeIngress *ingress) *Ingress {
 		ipAddressType = ipAddressTypeDualstack
 	}
 
+	sslPolicy := kubeIngress.getAnnotationsString(ingressSSLPolicyAnnotation, a.ingressDefaultSSLPolicy)
+	if _, ok := validSSLPolicy[sslPolicy]; !ok {
+		sslPolicy = a.ingressDefaultSSLPolicy
+	}
+
```
AlexanderYastrebov: This uses default policy if annotation is not set - that is expected. If annotation is set AND(BUT) policy is invalid it silently fallbacks to default if I read it correctly. Should it rather fail hard if it detects invalid policy?

szuecs:
:+1:

kubernetes/adapter.go
mikkeloscar: Problem of failing hard is we could break an existing ingress which is currently working for some clients, but then would break if misconfigured. We don't have a good signal to indicate that the ingress should continue to work in the previous configuration, but that the current configuration is broken.

I would like to work on better error reporting, but it's a bigger task which should be done separately.

I agree that if we merge this we can have hard to debug problems for users who misconfigure their ingress, but it's still better than the current state where the whole AWS account blows up.

mikkeloscar:
:+1:

PR #289, opened 2019-11-06
Make HTTP support opt-in for Network Load Balancers

mikkeloscar:
This adds a flag `--nlb-http-enabled` which when set
to true it will create an HTTP listener (port 80) for Network Load
Balancers. This value is disabled by default for security reasons since
NLB can't support HTTP -> HTTPS redirection.

This also improves the README by documenting the different features
available for the two different load balancer types supported.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ This information is used to manage AWS resources for each ingress objects of the
 - Automatic discovery of SSL certificates
 - Automatic forwarding of requests to all Worker Nodes, even with auto scaling
 - Automatic cleanup of unnecessary managed resources
-- Support for both [Application Load Balancers](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/introduction.html) and [Network Load Balancers](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/introduction.html).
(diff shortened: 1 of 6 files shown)


szuecs:
:+1:

aermakov-zalando:
Shouldn't this be configurable per ingress?

mikkeloscar:
> Shouldn't this be configurable per ingress?

Yes, we can do this but I would think of this as a bigger separate task considering ALB also doesn't support this per ingress at the moment.

I don't see a use case for this in Zalando so it's not high priority for us IMO. WDYT?

aermakov-zalando:
@mikkeloscar Some clients could be relying on the http->https redirect, I guess. But then it's better to break them since it's insecure, so this is probably fine.

mikkeloscar:
> @mikkeloscar Some clients could be relying on the http->https redirect, I guess. But then it's better to break them since it's insecure, so this is probably fine.

Note this change won't deploy NLBs by default. You have to opt-in to get an NLB. and http -> https redirect doesn't work for NLB anyway, so they would just go to the port 80 and not get redirected if they were depending on this and the ingress changed to NLB.

mikkeloscar:
:+1:

PR #302, opened 2020-01-10
Cleanup target groups from untargeted ASGs

### LINKED ISSUE #295: TG not cleaned up from ASGs where the filter no longer match.
If you change the `CUSTOM_FILTERS` value in a way such that an ASG which previously matched the filter no longer matches, then you end up in a situation where the ASG will not be managed by the controller, but it will still have a reference to the TG of the Load Balancer defined.

The controller should probably be able to clean up TGs which it owns from ASGs which it doesn't consider for attaching.

mikkeloscar:
This introduces logic for tracking ASGs which are considered "owned" based on ClusterID tag but are not configured to be targeted by the ALB.

This allows removing owned TargetGroups from ASGs which are no longer targeted because the `CUSTOM_FILTERS` configuration changed.

Fix #295 

## MIssing

* [x] Document this behavior in README


szuecs:
:+1:

mikkeloscar:
:+1:

PR #304, opened 2020-01-28
support skipper RouteGroups 

szuecs:
support skipper RouteGroups and create shared ALBs/NLBs and all annotations from ingress also in routegroups

TBD: change in client.patch() to change from `application/strategic-merge-patch+json` to  `application/merge-patch+json`, because the other is not supported for custom CRDs.


kubernetes/adapter.go
```diff
@@ -65,6 +68,7 @@ type Ingress struct {
 	SSLPolicy        string
 	IPAddressType    string
 	LoadBalancerType string
+	typ              string // "ingress" or "routegroup" to patch status based on that
```
mikkeloscar: maybe call it `resourceType` instead of this weird word? :)

mikkeloscar:
Generally it seems fine to me. Will wait for the patch change.

kubernetes/adapter.go
szuecs: fixed

szuecs: I changed the name from typ to resourceType.

szuecs:
I tested that the patch version is fine also for changing ALBs. So we do not get leftovers or such things in the status.

mikkeloscar:
:+1:

szuecs:
:+1:

PR #306, opened 2020-02-03
Add support for enabling/disabling HTTP2 (default true)

mikkeloscar:
This adds support for enabling/disabling http2 support an an Application Load Balancer via the annotation: `zalando.org/aws-load-balancer-http2: "false"`. The default is `true` like it was before where you could not change the setting.

Additionally I combined some of the `RouteGroup` and `Ingress` code to avoid duplication.


README.md
```diff
@@ -123,6 +124,7 @@ the individual Load Balancer types.
 | [Dualstack support][dualstack] | :heavy_check_mark: `--ip-addr-type=dualstack` | :heavy_multiplication_x: |
 | [Idle Timeout][idle_timeout] | :heavy_check_mark: `--idle-connection-timeout` | :heavy_multiplication_x: |
 | Custom Security Group | :heavy_check_mark: | :heavy_multiplication_x: |
+| HTTP/2 Support | :heavy_check_mark: | (not relevant) |
```
szuecs: maybe better to have a :weak_check_mark: , because ALBs convert H2 to HTTP/1.1 to call the backends

mikkeloscar: Made it :white_check_mark: instead of :heavy_check_mark: :)

kubernetes/adapter.go
```diff
@@ -291,26 +272,10 @@ func newIngressForKube(i *Ingress) *ingress {
 }
 
 func newRouteGroupForKube(i *Ingress) *routegroup {
-	shared := "true"
-
-	if !i.Shared {
-		shared = "false"
-	}
+	ing := newIngressForKube(i)
```
szuecs: this reads a bit weird, maybe better to refactor it into it's own function to prepare the Metadata.
wdyt?

szuecs:
Only one comment, everything else lgtm

kubernetes/adapter.go
mikkeloscar: Moved the metadata creation into a new `newMetadataForKube` function

szuecs:
:+1:

mikkeloscar:
:+1:

PR #314, opened 2020-02-24
RouteGroup CRD should be optional

### LINKED ISSUE #312: v0.10.0+ will fail to run if RouteGroups CRD is not installed
Recently tried a newer version of this, and found I had to revert because we do not have this CRD installed.

Should this be optional for those continuing to use traditional Ingress records?

szuecs:
fix #312 - make RouteGroup CRD optional and log a warning to notify the user about RouteGroups

### DIFF
--- a/kubernetes/adapter.go
+++ b/kubernetes/adapter.go
@@ -16,6 +16,7 @@ type Adapter struct {
 	ingressDefaultSecurityGroup    string
 	ingressDefaultSSLPolicy        string
 	ingressDefaultLoadBalancerType string
+	routeGroupSupport              bool
 }
 
 type ingressType int
@@ -122,6 +123,7 @@ func NewAdapter(config *Config, ingressClassFilters []string, ingressDefaultSecu
 		ingressDefaultSecurityGroup:    ingressDefaultSecurityGroup,
 		ingressDefaultSSLPolicy:        ingressDefaultSSLPolicy,
 		ingressDefaultLoadBalancerType: loadBalancerTypesAWSToIngress[ingressDefaultLoadBalancerType],
+		routeGroupSupport:              true,
 	}, nil
 }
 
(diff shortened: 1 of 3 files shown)


kubernetes/adapter.go
```diff
@@ -300,10 +301,16 @@ func (a *Adapter) IngressFiltersString() string {
 func (a *Adapter) ListResources() ([]*Ingress, error) {
 	ings, err := a.ListIngress()
 	if err != nil {
+		log.Errorf("Failed to list ingress: %v", err)
```
mikkeloscar: I would not log an error here since we return it on the line below and handle it from the caller.

kubernetes/client.go
```diff
@@ -84,16 +87,20 @@ func (c *simpleClient) get(resource string) (io.ReadCloser, error) {
...
-		resp.Body.Close()
-		return nil, err
+
+	if resp.StatusCode == http.StatusOK {
+		return resp.Body, nil
 	}
-	return resp.Body, nil
+	defer resp.Body.Close()
+
+	if resp.StatusCode == int(http.StatusNotFound) {
```
mikkeloscar: nit: it's already an `int`

kubernetes/adapter.go
szuecs: Ỳep duplication of logs, dropped

kubernetes/client.go
szuecs: yeah left over of some weird debugging

kubernetes/adapter.go
```diff
@@ -304,6 +305,11 @@ func (a *Adapter) ListResources() ([]*Ingress, error) {
 	}
 	rgs, err := a.ListRoutegroups()
 	if err != nil {
+		log.Warnf("Failed to list RouteGroups: %v, to get more information https://opensource.zalando.com/skipper/kubernetes/routegroups/#routegroups", err)
```
mikkeloscar: I'm afraid this will be quite spammy in the logs as it will happen every sync interval. IMO it should be ok to run without RouteGroup support and you don't want to see a lot of logs telling the same thing if you can't turn it off.

Either we can add a flag to disable route groups completely, or we can detect the first time we try to list routegroups and e.g. set a field `a.routeGroupSupport = false` and only log `if a.routeGroupSupport`. WDYT?

szuecs: From 9 to 10 lines per interval is too spammy?

szuecs: The problem with storing the value is that after CRD apply it does not work, but you would have to delete the pod to retry.

mikkeloscar: You could still try to list the routes every interval (and reset the flag). I just think it doesn't bring any value to the user to log a warning that they can't do anything about (apart from deploying the CRD which they may not need/want).

szuecs: This makes sense!
I added as suggested the check

szuecs:
:+1:

mikkeloscar:
:+1:

PR #315, opened 2020-02-26
feature: support cluster.local 

### LINKED ISSUE #308: omit LB creation for cluster.local domain (was: adding a load balancer type none)
In order to support [cluster internal east-west routing](https://opensource.zalando.com/skipper/kubernetes/east-west-usage/) without having an assigned load balancer we should create a `zalando.org/aws-load-balancer-type: none`, such that users can opt-out the load balancer creation and external-dns would not create a DNS name for it.

Why not using class to opt-out selection of this ingress-controller? 
Because you might want to run it as default ingress controller such that it will pick up the ingress anyways.

szuecs:
fix #308
- feature: add cluster.local to not create an ALB/NLB
- feature: update ingress/routegroup, in case you switch from public to only cluster.local host rules

### DIFF
--- a/kubernetes/adapter.go
+++ b/kubernetes/adapter.go
@@ -142,7 +142,7 @@ func (a *Adapter) newIngressFromKube(kubeIngress *ingress) *Ingress {
 	}
 
 	for _, rule := range kubeIngress.Spec.Rules {
-		if rule.Host != "" && !strings.HasSuffix(rule.Host, a.clusterLocalDomain) {
+		if rule.Host != "" && (a.clusterLocalDomain == "" || !strings.HasSuffix(rule.Host, a.clusterLocalDomain)) {
 			hostnames = append(hostnames, rule.Host)
 		}
 	}
@@ -170,7 +170,7 @@ func (a *Adapter) newIngressFromRouteGroup(rg *routegroup) *Ingress {
 	}
 
 	for _, host := range rg.Spec.Hosts {
-		if host != "" && !strings.HasSuffix(host, a.clusterLocalDomain) {
+		if host != "" && (a.clusterLocalDomain == "" || !strings.HasSuffix(host, a.clusterLocalDomain)) {
 			hostnames = append(hostnames, host)
 		}
 	}
--- a/kubernetes/adapter.go
+++ b/kubernetes/adapter.go
@@ -342,19 +342,13 @@ func (a *Adapter) ListIngress() ([]*Ingress, error) {
 			ingressClass := getAnnotationsString(ingress.Metadata.Annotations, ingressClassAnnotation, "")
 			for _, v := range a.ingressFilters {
 				if v == ingressClass {
-					ing := a.newIngressFromKube(ingress)
-					if ing != nil {
-						ret = append(ret, ing)
-					}
+					ret = append(ret, a.newIngressFromKube(ingress))
 				}
 			}
 		}
 	} else {
 		for _, ingress := range il.Items {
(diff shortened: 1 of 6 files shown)


kubernetes/ingress.go
```diff
@@ -71,6 +71,7 @@ const (
 	ingressLoadBalancerTypeAnnotation = "zalando.org/aws-load-balancer-type"
 	ingressHTTP2Annotation            = "zalando.org/aws-load-balancer-http2"
 	ingressClassAnnotation            = "kubernetes.io/ingress.class"
+	clusterLocalDomain                = ".cluster.local"
```
mikkeloscar: Let's make this a flag because users might have a different cluster domain depending on their setup.

mikkeloscar:
@szuecs Since we currently orchestrate update of the ingress status from `worker.go` I think it may be simpler to just track a "special" clusterLocal=true `loadBalancer` there such that we can handle the case where the hostname should be removed from the ingress status.

So my idea would be to create a loadBalancer here: https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/e18e972d3eb46c1bb2fe1abcf358b059db959157/worker.go#L345-L382 with a field `clusterLocal: true` and add all the relevant ingresses there. Then we can later ensure the ingress status is cleared here: https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/e18e972d3eb46c1bb2fe1abcf358b059db959157/worker.go#L270-L283

kubernetes/adapter.go
```diff
@@ -332,19 +338,23 @@ func (a *Adapter) ListIngress() ([]*Ingress, error) {
...
 	var ret []*Ingress
 	if len(a.ingressFilters) > 0 {
-		ret = make([]*Ingress, 0)
 		for _, ingress := range il.Items {
 			ingressClass := getAnnotationsString(ingress.Metadata.Annotations, ingressClassAnnotation, "")
 			for _, v := range a.ingressFilters {
 				if v == ingressClass {
-					ret = append(ret, a.newIngressFromKube(ingress))
+					ing := a.newIngressFromKube(ingress)
+					if ing != nil {
```
mikkeloscar: This nil check is no longer needed as we don't return `nil` anymore

worker.go
```diff
@@ -321,8 +336,26 @@ func getAllLoadBalancers(certTTL time.Duration, stacks []*aws.Stack) []*loadBala
...
+
+	return &loadBalancer{
+		clusterLocal: true,
+	}
+}
+
 func matchIngressesToLoadBalancers(loadBalancers []*loadBalancer, certs CertificatesFinder, certsPerALB int, ingresses []*kubernetes.Ingress) []*loadBalancer {
 	for _, ingress := range ingresses {
+		if ingress.ClusterLocal {
+			lb := getClusterLocal(loadBalancers)
```
mikkeloscar: Since the `loadBalancers` we pass to `matchIngressesToLoadBalancers` are only build based on current stacks, it will never contain a `ClusterLocal` LB. I think it's maybe simpler to just define a `ClusterLocal` `loadBalancer` before the loop and then just add ingresses to that instead of the `getClusterLocal` function which will go through all load balancers and return a new one always anyway.

kubernetes/adapter.go
szuecs: good catch, much cleaner to read :)

worker.go
szuecs: changed as suggested

mikkeloscar:
:+1:

szuecs:
:+1:

PR #327, opened 2020-04-01
Add flag to disable instrumented http client

### LINKED ISSUE #309: Extreme slowdown since upgrading to v0.9.7 (5 minutes)
When I run on 0.9.6 I get all debug messages with `--debug` flag.

But starting from `v0.9.7` it stops, and I only see:
```
public-ingress-controller-58d54554c7-rknd6 controller time="2020-02-13T09:07:06Z" level=info msg="starting /kube-ingress-aws-controller"
```

I don't see anything weird in https://github.com/zalando-incubator/kube-ingress-aws-controller/compare/v0.9.6...v0.9.7

ruudk:
Fixes #309

See https://github.com/zalando-incubator/kube-ingress-aws-controller/issues/309#issuecomment-607275390

### DIFF
--- a/aws/adapter.go
+++ b/aws/adapter.go
@@ -151,12 +151,14 @@ var (
 	}
 )
 
-func newConfigProvider(debug bool) client.ConfigProvider {
+func newConfigProvider(debug, disableInstrumentedHttpClient bool) client.ConfigProvider {
 	cfg := aws.NewConfig().WithMaxRetries(3)
 	if debug {
 		cfg = cfg.WithLogLevel(aws.LogDebugWithRequestErrors)
 	}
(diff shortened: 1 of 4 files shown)


ruudk:
I tested this and it works without issues now 🎉

mikkeloscar:
:+1:

szuecs:
@mikkeloscar what do you think if we use https://godoc.org/github.com/zalando/skipper/net#Client instead of the instrumented client?
Not sure about the current data we get from the instrumented client.

szuecs:
:+1:

mikkeloscar:
@szuecs using the other client wouldn't solve this particular problem as far as I understand because it would trigger the same issue where aws-sdk doesn't use a lower timeout for going to the metadata endpoint.

PR #329, opened 2020-04-01
Remove tabs from logs

ruudk:
It does not make sense to log tabs. This thing runs in Kubernetes and when viewing those logs, the tabs are printed as `\t`.


ruudk:
@mikkeloscar Is there a way to disable the auto posting of comments by @coveralls? It's sooo annoying to see that you have a notification, and thinking that someone replied, and then get the disappointed from this wining bot. It can also log it's results in Github Checks, right?

mikkeloscar:
:+1:

szuecs:
:+1:

PR #341, opened 2020-04-08
LB-WAF association from ingress

aryszka:
- [x] apply WAF Web ACL parameter
- [x] detect updates
- [x] use separate LB for ingresses with and without WAF
- [x] collapse identical stacks after WAF was removed
- [x] fix global WAF if necessary
- [x] conventional abbreviation casing
- [x] tests
- [x] docs

### DIFF
--- a/kubernetes/adapter.go
+++ b/kubernetes/adapter.go
@@ -89,7 +89,7 @@ type Ingress struct {
 	SSLPolicy        string
 	IPAddressType    string
 	LoadBalancerType string
-	WafWebACLId      string
+	WAFWebACLId      string
 	Hostnames        []string
 	resourceType     ingressType
 }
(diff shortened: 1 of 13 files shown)


kubernetes/adapter.go
```diff
@@ -89,7 +89,7 @@ type Ingress struct {
 	SSLPolicy        string
 	IPAddressType    string
 	LoadBalancerType string
-	WafWebACLId      string
+	WAFWebACLId      string
```
mikkeloscar, line 92: nit: Should the name be `WAFWebACLID` (`Id -> ID`) to follow the go conventions?

worker.go
```diff
@@ -92,7 +92,7 @@ func (l *loadBalancer) AddIngress(certificateARNs []string, ingress *kubernetes.
 		l.securityGroup != ingress.SecurityGroup ||
 		l.sslPolicy != ingress.SSLPolicy ||
 		l.loadBalancerType != ingress.LoadBalancerType ||
-		l.http2 != ingress.HTTP2 || l.wafWebACLId != ingress.WafWebACLId {
+		l.http2 != ingress.HTTP2 {
```
mikkeloscar, line 95: This `if` has the function of grouping ingresses on load balancers based on all the parameters. If you don't remove the `wafWebACLId` check, then there should not be a need for the  `groupLBsByCurrentWAF()` function below as it's effectively the same thing from what I understand, just implemented in a different way.

aryszka: there was a reason during an interim state for making this change. Basically changes were not properly handled, but I need to check in detail. If possible, then I'll revert this change to make it analogous to the rest of these properites.

aryszka: done

mikkeloscar:
:+1:

aryszka:
:+1:

Issue #320, opened 2020-03-27
Support for AWS WAF v2


janario:
AWS has build a new version with new feature of WAF, in this case V2

Evaluating the current cloud formation template, when an waf id is provided, it generates something like:

```
WAFAssociation": {
    "Type": "AWS::WAFRegional::WebACLAssociation",
    "Properties": {
        "ResourceArn": {
            "Ref": "LB"
        },
        "WebACLId": "<my-waf-v1-id>"
    },
    "Metadata": {
        "AWS::CloudFormation::Designer": {
            "id": "..."
        }
    }
},
```

Which should be changed in case of WAF v2, something like:
```
"WAF2WAC1KPI3": {
    "Type": "AWS::WAFv2::WebACLAssociation",
    "Properties": {
        "ResourceArn": {
            "Ref": "LB"
        },
        "WebACLId": "<my-waf-v2-id>"
    },
    "Metadata": {
        "AWS::CloudFormation::Designer": {
            "id": "..."
        }
    }
}
```

It would be nice to have a different flag or something to support the waf version 2

* `-aws-waf2-web-acl-id` ?

szuecs:
It should also be possible to set this per ingress, so we need an annotation, too

janario:
Created an MR https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/342

Turns out for waf v2 cloud formation expects an ARN, since the parameter value will be complete different added a condition where when it starts with `arn:aws:wafv2:` it will use waf v2

-aws-waf-web-acl-id

janario:
Closed by #342

janario:
@aryszka 
@mikkeloscar 
@szuecs 

Hi, Can we include this issue or the PR https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/342 in the release?

https://github.com/zalando-incubator/kube-ingress-aws-controller/releases/tag/v0.10.12

Just notice v0.10.12 already contains it
```
$ docker run -it registry.opensource.zalan.do/teapot/kube-ingress-aws-controller:v0.10.12 --help | grep waf
  --aws-waf-web-acl-id=""        WAF
                                 arn:aws:wafv2:<region>:<account>:regional/webacl/<name>/<id>

```

szuecs:
@janario it is already released  in https://github.com/zalando-incubator/kube-ingress-aws-controller/releases/tag/v0.10.11
We have release automation that release every merge to master.

[...]

Issue #331, opened 2020-04-01
ValidationError: Parameters: [LoadBalancerWafWebACLIdParameter] do not exist in the template


ruudk:
After upgrading the controller from `v0.9.6` to latest I'm getting:
```
private-ingress-controller-5b444bb4d4-4hvbn controller 2020/04/01 17:49:54 DEBUG: Validate Response cloudformation/UpdateStack failed, attempt 0/3, error ValidationError: Parameters: [LoadBalancerWafWebACLIdParameter] do not exist in the template
private-ingress-controller-5b444bb4d4-4hvbn controller 	status code: 400, request id: 70b42810-964d-44d3-bdec-15ec6834026f
private-ingress-controller-5b444bb4d4-4hvbn controller time="2020-04-01T17:49:54Z" level=error msg="updateStack(map[\"arn:aws:acm:eu-west-1:355892651979:certificate/b9a91d7b-74a3-4534-a9c1-3fa718c7bb87\":\"0001-01-01 00:00:00 +0000 UTC\" \"arn:aws:acm:eu-west-1:355892651979:certificate/c20bb770-16ec-408f-8d1f-be3170277d0a\":\"0001-01-01 00:00:00 +0000 UTC\"]) failed: ValidationError: Parameters: [LoadBalancerWafWebACLIdParameter] do not exist in the template\n\tstatus code: 400, request id: 70b42810-964d-44d3-bdec-15ec6834026f"
```

@aryszka Did you encounter the same issue in https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/326?

mikkeloscar:
@ruudk yes, we're working on a fix today!

ruudk:
@mikkeloscar Any news on this? :)

mikkeloscar:
Should be fixed by #342 from what i understand.

ruudk:
This problem still exists with the latest release v0.10.12
```
private-ingress-controller-768684f8fb-wr2kg controller 2020/04/21 08:16:44 DEBUG: Validate Response cloudformation/UpdateStack failed, attempt 0/3, error ValidationError: Parameters: [LoadBalancerWAFWebACLIDParameter] must have values
private-ingress-controller-768684f8fb-wr2kg controller 	status code: 400, request id: 699eaf95-2cdb-41e0-b9e7-fe49ba0e18c0
[...]

PR #342, opened 2020-04-08
#320 Added support for WAF v2 based on waf parameter

janario:
Related to https://github.com/zalando-incubator/kube-ingress-aws-controller/issues/320

Added support for WAF v2, based on the parameter when it is an ARN of WAF2 it will create the right template

### DIFF
--- a/aws/cf_template.go
+++ b/aws/cf_template.go
@@ -275,7 +275,7 @@ func generateTemplate(spec *stackSpec) (string, error) {
 
 	if spec.loadbalancerType == LoadBalancerTypeApplication && spec.wafWebAclId != "" {
 		if strings.HasPrefix(spec.wafWebAclId, "arn:aws:wafv2") {
-			template.AddResource("WAF2WAC1KPI3", &cloudformation.WAFv2WebACLAssociation{
+			template.AddResource("WAFAssociation", &cloudformation.WAFv2WebACLAssociation{
 				ResourceArn: cloudformation.Ref("LB").String(),
 				WebACLArn:   cloudformation.String(spec.wafWebAclId),
 			})
--- a/aws/cf_template.go
+++ b/aws/cf_template.go
@@ -3,6 +3,7 @@ package aws
 import (
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	"crypto/sha256"
 	"sort"
@@ -81,6 +82,11 @@ func generateTemplate(spec *stackSpec) (string, error) {
 			Type:        "String",
 			Description: "H2 Enabled",
 			Default:     "true",
+		},
+		parameterLoadBalancerWafWebACLIdParameter: &cloudformation.Parameter{
+			Type:        "String",
+			Description: "WAF Id or ARN",
+			Default:     "",
 		},
 	}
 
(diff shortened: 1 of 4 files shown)


aws/cf_template.go
```diff
@@ -268,10 +274,17 @@ func generateTemplate(spec *stackSpec) (string, error) {
...
-		template.AddResource("WAFAssociation", &cloudformation.WAFRegionalWebACLAssociation{
-			ResourceArn: cloudformation.Ref("LB").String(),
-			WebACLID:    cloudformation.String(spec.wafWebAclId),
-		})
+		if strings.HasPrefix(spec.wafWebAclId, "arn:aws:wafv2") {
+			template.AddResource("WAF2WAC1KPI3", &cloudformation.WAFv2WebACLAssociation{
+				ResourceArn: cloudformation.Ref("LB").String(),
+				WebACLArn:   cloudformation.String(spec.wafWebAclId),
+			})
+		} else {
```
szuecs, line 282: What if we have no waf ID specified?
We should omit creating one as far as I understand.
Related https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/341

janario: It won't get into the first if

> if spec.loadbalancerType == LoadBalancerTypeApplication && spec.wafWebAclId != "" {

* no waf, out of the entire if
* waf arn v2, template for v2
* waf id (v1), template for v1 as it used to be

---
But depending if #341 gets merged first this one has to be changed

waf2 (I suppose this)
> WebACLArn:    cloudformation.Ref(parameterLoadBalancerWAFWebACLIdParameter).String(),

waf1
> WebACLID:    cloudformation.Ref(parameterLoadBalancerWAFWebACLIdParameter).String(),

aws/cf_template.go
```diff
@@ -268,10 +274,17 @@ func generateTemplate(spec *stackSpec) (string, error) {
 	})
 
 	if spec.loadbalancerType == LoadBalancerTypeApplication && spec.wafWebAclId != "" {
-		template.AddResource("WAFAssociation", &cloudformation.WAFRegionalWebACLAssociation{
-			ResourceArn: cloudformation.Ref("LB").String(),
-			WebACLID:    cloudformation.String(spec.wafWebAclId),
-		})
+		if strings.HasPrefix(spec.wafWebAclId, "arn:aws:wafv2") {
+			template.AddResource("WAF2WAC1KPI3", &cloudformation.WAFv2WebACLAssociation{
```
mikkeloscar, line 278: What does this name mean? `WAF2WAC1KPI3` can we use something more human readable? :)

janario: Good point

I got it from the visual template, but it can be renamed

Changed it to the same as from v1 `WAFAssociation`

mikkeloscar:
@janario If you can just fix the DCO check (signoff your commit) then it lgtm.

janario:
> @janario If you can just fix the DCO check (signoff your commit) then it lgtm.

@mikkeloscar thanks for the reminder I forgot the signoff 

Fixed it :+1:

mikkeloscar:
:+1:

aryszka:
:+1:

aryszka:
@janario thanks for the contribution!

aws/cf_template.go
ruudk: I think this merge went wrong, see https://github.com/zalando-incubator/kube-ingress-aws-controller/issues/331#issuecomment-617031946

@janario @aryszka @mikkeloscar @szuecs

aws/cf_template.go
```diff
@@ -82,6 +83,11 @@ func generateTemplate(spec *stackSpec) (string, error) {
 			Description: "H2 Enabled",
 			Default:     "true",
 		},
+		parameterLoadBalancerWafWebACLIdParameter: &cloudformation.Parameter{
+			Type:        "String",
+			Description: "WAF Id or ARN",
+			Default:     "",
```
szuecs, line 89: This is what I expect to be a problem mentioned in https://github.com/zalando-incubator/kube-ingress-aws-controller/issues/331
Could it be that this ends in the cf stack?

janario: I noticed the error while I was testing in my cluster

It looked like master was broken, I fixed in order to unblock my test

Sorry, I should have pointed it here

janario: I mean, it failed with an error even when I didn't provide an WAF config

Issue #339, opened 2020-04-06
Update ALB Security Policy instead of creating a new ALB


adri:
Hi 👋 

my goal is to update the SSL Policy config of an ALB managed by kube-ingress-aws-controller v0.9.6.

In my tests, I see that a new CloudFormation stack is created when I set the default `ssl-policy` on the ingress controller deployment. When creating a new CloudFormation stack, also a new load balancer is created and external-dns moves over the Route53 entries to the new load balancer. 

However we had issues in the past where not all DNS caches are updated and the old load balancer was removed. Is there a way to update the CloudFormation stack to update the Security  Policy to avoid changing DNS entries?

I think this line decides to not match the ingress with the load balancer:
https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/master/worker.go#L93
Is that on purpose? Is there another way to set the Security Policy without potential downtime?

szuecs:
SSLpolicy can’t be in general changed in place.
Let’s say you have 20 ingress and you change the policy on one....
I think the best would be to have longer ttl value for cf stack deletion. We already have a ttl which could be increased.
Maybe you can check if you already can do that. Or @mikkeloscar can check and answer 😇

adri:
hi @szuecs, thanks! That stack TTL sounds like a good idea in any case. 

Looks like it's added in this [commit](https://github.com/zalando-incubator/kube-ingress-aws-controller/commit/200d71a59104dd9449679f0c07efab369c58d314), however I don't see a way to modify it from outside. The default is set to 5 minutes, but the commit messages says 1h 🤔 
I'd set that to 48 hours to be sure most DNS caches are updated. Do you imagine a PR or making this configurable?

szuecs:
You are right it’s 5 minutes right now https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/master/aws/adapter.go#L83, because we tested it and forget to change it.
A PR can be done if you check for example the idle connection timeout in https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/master/aws/adapter.go#L265 and the worker.go file.
Can you create a PR, please?

szuecs:
@mikkeloscar maybe you have time to fix it?

adri:
I tested with setting the stack termination protection, but although it was enabled the stack was deleted after 30 minutes. So that's not an option :)

[...]

PR #344, opened 2020-04-09
Change default cert TTL to 1 hour

ruudk:
As discussed in https://github.com/zalando-incubator/kube-ingress-aws-controller/issues/339

There is already a way for us to increase the TTL:
```go
kingpin.Flag("cert-ttl-timeout", "sets the timeout of how long a certificate is kept on an old ALB to be decommissioned.").
   Default(defaultCertTTL).DurationVar(&certTTL)
```

This TTL is used to remove the old LB stack. It was set to 30 minutes and this PR will change it to 1 hour.

On our deployment, we will further increase this to 24 hours.

### DIFF
--- a/controller.go
+++ b/controller.go
@@ -25,7 +25,7 @@ const (
 	defaultDisableSNISupport      = "false"
 	defaultInstrumentedHttpClient = "false"
 	defaultHTTPRedirectToHTTPS    = "false"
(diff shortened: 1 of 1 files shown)


szuecs:
:+1:

mikkeloscar:
:+1:

PR #354, opened 2020-06-02
Add support for networking.k8s.io/v1beta1

mikkeloscar:
Adds support for listing and updating ingresses via the newer `networking.k8s.io/v1beta1` apiVersion. This also becomes the default but can be reverted back to `extensions/v1beta1` via the flag `--ingress-api-version`. The motivation for making it the default is that it was introduced already in Kubernetes v1.14.


szuecs:
@mikkeloscar great!
Please increase the version in https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/master/VERSION and add a note to https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/master/README.md#upgrade. Clusters older than 1.14 have to do something.

mikkeloscar:
@szuecs I have update version and README

szuecs:
:+1:

mikkeloscar:
:+1:

PR #363, opened 2020-07-17
UpdateStack(): preserve the owner tag

aermakov-zalando:
Since we update on startup, the controller will erase the owner tag on every restart, which triggers a recreation of every non-shared ELB. This causes issues if the new instances don't have enough capacity to deal with the influx of requests, and creates churn that's not even necessary. Fix by preserving the ownership tag correctly.


szuecs:
:+1:

aermakov-zalando:
👍

PR #371, opened 2020-09-08
Add support for setting additional stack tags

mikkeloscar:
Adds a flag `--additional-stack-tags key=val` which allows setting additional custom tags on the CF stacks managed by the controller.

This can be useful if your infrastructure relies on additional tags e.g. for cost reporting or permissions. Zalando use case is to add `InfrastructureComponent=true` tags such that we can prevent non-admin users from touching resources with those tags.


szuecs:
:+1:

mikkeloscar:
:+1:

PR #372, opened 2020-09-16
Make deregistration delay timeout configurable

aermakov-zalando:
Allow configuring the [deregistration delay](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/load-balancer-target-groups.html#deregistration-delay) via `--deregistration-delay-timeout`. The default is set to 5 minutes to match AWS.


mikkeloscar:
:+1:

aermakov-zalando:
👍

PR #381, opened 2020-11-18
UpdateStack(): don't reset the security group

aermakov-zalando:
`CreateStack` and `UpdateStack` used different values for the SG parameter, causing it to reset the configuration. Fix by making `UpdateStack` use the SG from the ingress if set, not the default one.


szuecs:
:+1:

aryszka:
:+1:

aryszka:
:+1:

szuecs:
:+1:

PR #389, opened 2021-01-04
Fix detaching of target groups

mikkeloscar:
In some cases the Autoscaling groups can end up being configured with target groups that no longer exists. If that happens we have logic for cleaning up those target groups to prevent other autoscaling group changes from being blocked.

This fixes the logic slightly such that it doesn't try to get the tags of non-existing target groups which would fail and thus prevent the controller from detaching the non-existing target groups.


szuecs:
:+1:

mikkeloscar:
:+1:

PR #401, opened 2021-02-10
Don't set healthcheck timeout on NLB

mikkeloscar:
For some reason custom healthcheck timeout is not supported when target group uses TCP protocol. This is the error shown in Cloudformation when creating an NLB:

```
Custom health check timeouts are not supported for health checks for target groups with the TCP protocol (Service: AmazonElasticLoadBalancing; Status Code: 400; Error Code: InvalidConfigurationRequest; Request ID: xxx; Proxy: null)
```

Fix it by only setting custom healthcheck timeout when `protocol != TCP`

### DIFF
--- a/aws/cf_template.go
+++ b/aws/cf_template.go
@@ -277,18 +277,23 @@ func generateTemplate(spec *stackSpec) (string, error) {
 			Value: cloudformation.String(fmt.Sprintf("%d", spec.deregistrationDelayTimeoutSeconds)),
 		},
 	}
-	template.AddResource("TG", &cloudformation.ElasticLoadBalancingV2TargetGroup{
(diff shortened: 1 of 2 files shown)


jonathanbeber (commented):
The changes look good to me, I just wonder If we could/should add some unit tests for that. It seems easy to change this code and remove/forget about this condition. I'm working on this package's tests for ALB host headers rule and if you want I could give it a quick try today.

What do you think?

mikkeloscar:
@jonathanbeber you're welcome to add tests. Otherwise I can also do it next week :)

jonathanbeber:
@mikkeloscar will do it.

jonathanbeber:
:+1:

mikkeloscar:
:+1:

PR #400, opened 2021-02-09
ALB Listener rule to deny internal domain requests

jonathanbeber:
This commit addresses #402. It adds the following new configuration
flags:
  - internal-domains
  - deny-internal-domains
  - deny-internal-domains-response
  - deny-internal-domains-response-content-type
  - deny-internal-domains-response-status-code

The main flag is the `deny-internal-domains`, a boolean that enables and
disables the feature. The other flags allow to set the internal domains
that will be protected as well as the response that will be returned.
Defaults are also set: an HTTP `text/plain` `401 - Unauthorized`
response for the `*.ingress.cluster.local` domain.

The ALB cloudformation template will add a
[`AWS::ElasticLoadBalancingV2::ListenerRule`][0] resource, configured
with the [condition][1] values from the `internal-domains` and the
[action `fixedresponseconfig`][2] with the respective response
`deny-internal-domains-response` flags.

[0]: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticloadbalancingv2-listenerrule.html
[1]: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-elasticloadbalancingv2-listenerrule-hostheaderconfig.html
[2]: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-elasticloadbalancingv2-listenerrule-action.html#cfn-elasticloadbalancingv2-listenerrule-action-fixedresponseconfig

Signed-off-by: Jonathan Juares Beber <jonathanbeber@gmail.com>

PR #404, opened 2021-02-16
Apply rule creation just on ALBs

jonathanbeber:
The #400 PR added rules to deny internal traffic, and in its docs it
even stated that NLBs do not support this feature. But the code didn't
check for the load balancer type. This commit adds the logic to apply
the rule creation just for ALBs.

### DIFF
--- a/aws/cf_template.go
+++ b/aws/cf_template.go
@@ -151,7 +151,8 @@ func generateTemplate(spec *stackSpec) (string, error) {
 			Port:            cloudformation.Integer(80),
 			Protocol:        cloudformation.String(protocol),
 		})
-		if spec.denyInternalDomains {
+		// Just ALBs support Rules
+		if spec.denyInternalDomains && spec.loadbalancerType == LoadBalancerTypeApplication {
 			template.AddResource(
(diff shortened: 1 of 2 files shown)


aws/cf_template.go
```diff
@@ -151,7 +151,8 @@ func generateTemplate(spec *stackSpec) (string, error) {
 			Port:            cloudformation.Integer(80),
 			Protocol:        cloudformation.String(protocol),
 		})
-		if spec.denyInternalDomains {
+		// NLBs do not support Rules
+		if spec.denyInternalDomains && protocol != "TCP" {
```
mikkeloscar, line 155: This should not be based on protocol but on the load balancer type, see how we check it elsewhere.

jonathanbeber:
:+1:

aws/cf_template.go
jonathanbeber: good one, it actually makes even more sense to allow only ALBs instead of filtering out NLBs.

jonathanbeber:
:+1:

mikkeloscar:
:+1:

PR #411, opened 2021-03-25
Add the ability to use an HTTPS target

jhuntwork:
This adds a boolean flag called `target-https` which will set the
protocol used on a target group to HTTPS.

**Motivation**

AWS VPC networking does prevent external targets from sniffing the wire,
but if the target is a proxy listening on the hostNetwork, any other pod
deployed to that node that is also using the hostNetwork can read the traffic
unencrypted. In cases where end-to-end encryption is expected or
desirable, this allows for that option.

### DIFF
--- a/aws/cf_template.go
+++ b/aws/cf_template.go
@@ -24,6 +24,9 @@ const (
 	//
 	// [0]: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticloadbalancingv2-listenerrule.html#cfn-elasticloadbalancingv2-listenerrule-priority
 	internalTrafficDenyRulePriority int64 = 1
+
+	httpProtocol  = "HTTP"
+	httpsProtocol = "HTTPS"
 )
 
 func hashARNs(certARNs []string) []byte {
@@ -112,11 +115,13 @@ func generateTemplate(spec *stackSpec) (string, error) {
 		}
 	}
 
-	protocol := "HTTP"
-	tlsProtocol := "HTTPS"
+	protocol := httpProtocol
+	tlsProtocol := httpsProtocol
 	if spec.loadbalancerType == LoadBalancerTypeNetwork {
 		protocol = "TCP"
 		tlsProtocol = "TLS"
+	} else if spec.targetHTTPS {
+		protocol = httpsProtocol
 	}
 
 	if spec.loadbalancerType == LoadBalancerTypeApplication && spec.httpRedirectToHTTPS {
@@ -125,7 +130,7 @@ func generateTemplate(spec *stackSpec) (string, error) {
 				{
 					Type: cloudformation.String("redirect"),
 					RedirectConfig: &cloudformation.ElasticLoadBalancingV2ListenerRedirectConfig{
-						Protocol:   cloudformation.String("HTTPS"),
+						Protocol:   cloudformation.String(httpsProtocol),
 						Port:       cloudformation.String("443"),
 						Host:       cloudformation.String("#{host}"),
 						Path:       cloudformation.String("/#{path}"),
@@ -136,7 +141,7 @@ func generateTemplate(spec *stackSpec) (string, error) {
 			},
 			LoadBalancerArn: cloudformation.Ref("LB").String(),
 			Port:            cloudformation.Integer(80),
-			Protocol:        cloudformation.String("HTTP"),
+			Protocol:        cloudformation.String(httpProtocol),
 		})
 	} else if spec.loadbalancerType == LoadBalancerTypeApplication || spec.nlbHTTPEnabled {
 		listenerName := "HTTPListener"
@@ -325,7 +330,7 @@ func generateTemplate(spec *stackSpec) (string, error) {
 		HealthCheckIntervalSeconds: cloudformation.Ref(parameterTargetGroupHealthCheckIntervalParameter).Integer(),
 		HealthCheckPath:            cloudformation.Ref(parameterTargetGroupHealthCheckPathParameter).String(),
 		HealthCheckPort:            cloudformation.Ref(parameterTargetGroupHealthCheckPortParameter).String(),
-		HealthCheckProtocol:        cloudformation.String("HTTP"),
+		HealthCheckProtocol:        cloudformation.String(protocol),
 		Port:                       cloudformation.Ref(parameterTargetTargetPortParameter).Integer(),
 		Protocol:                   cloudformation.String(protocol),
 		VPCID:                      cloudformation.Ref(parameterTargetGroupVPCIDParameter).String(),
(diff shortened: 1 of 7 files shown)


jhuntwork:
Just as a note: I believe this will always be a non-validated cert on the target. Self-signed certs work, and appear to be acceptable. AWS documentation supports that, here: https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-target-groups.html#target-group-routing-configuration

> Because the load balancer is in a virtual private cloud (VPC), traffic between the load balancer and the targets is authenticated at the packet level, so it is not at risk of man-in-the-middle attacks or spoofing even if the certificates on the targets are not valid. 

And further insight into AWS thinking about why they feel this is reasonable is here: https://kevin.burke.dev/kevin/aws-alb-validation-tls-reply/

jhuntwork:
Will do!

jhuntwork:
Still a small drop in coverage, mostly due to the new function in `adapter.go`. There don't seem to be any existing similar tests for the Adapter struct. Should I add new ones?

mikkeloscar:
:+1:

aws/cf_template.go
```diff
@@ -149,7 +155,7 @@ func generateTemplate(spec *stackSpec) (string, error) {
 			},
 			LoadBalancerArn: cloudformation.Ref("LB").String(),
 			Port:            cloudformation.Integer(80),
-			Protocol:        cloudformation.String(protocol),
+			Protocol:        cloudformation.String(httpProtocol),
```
jonathanbeber, line 158: It will make NLB have the `HTTP` protocol and it's not supported.

We should have it tested too. NLB's `HTTPListener` should be checked to be using `TCP`, not `HTTP`.

jonathanbeber (changes requested):

aws/cf_template.go
jonathanbeber: I think the following might be sufficient:
```diff
diff --git aws/cf_template_test.go aws/cf_template_test.go
index 16860e9..fc60f1c 100644
--- aws/cf_template_test.go
+++ aws/cf_template_test.go
@@ -161,6 +161,7 @@ func TestGenerateTemplate(t *testing.T) {
                        validate: func(t *testing.T, template *cloudformation.Template) {
                                require.NotNil(t, template.Resources["HTTPListener"])
                                listener := template.Resources["HTTPListener"].Properties.(*cloudformation.ElasticLoadBalancingV2Listener)
+                               require.Equal(t, listener.Protocol, cloudformation.String("TCP"))
                                require.Len(t, *listener.DefaultActions, 1)
                                redirectConfig := []cloudformation.ElasticLoadBalancingV2ListenerAction(*listener.DefaultActions)[0].RedirectConfig
                                require.Nil(t, redirectConfig)

```

mikkeloscar: Good point @jonathanbeber 

Another thing is that you can't change the protocol of the health check on an NLB, so we actually need to handle that case when deciding which LBs to create, or at least document that changing the flag when you already have existing NLBs is not supported

jhuntwork: Thanks, good catch. Adjusted both.

jhuntwork: > Another thing is that you can't change the protocol of the health check on an NLB, so we actually need to handle that case when deciding which LBs to create, or at least document that changing the flag when you already have existing NLBs is not supported

I believe we can make it simpler by just having this affect ALBs. That was essentially my intention in the first place. I'll push another change up to reflect that, as well as specific tests to demonstrate NLBs are not impacted.

aws/cf_template_test.go
```diff
@@ -355,6 +355,50 @@ func TestGenerateTemplate(t *testing.T) {
...
+			validate: func(t *testing.T, template *cloudformation.Template) {
+				tg, ok := template.Resources["TG"].Properties.(*cloudformation.ElasticLoadBalancingV2TargetGroup)
+				require.True(t, ok, "couldn't convert resource to ElasticLoadBalancingV2TargetGroup")
+				listener := template.Resources["HTTPListener"].Properties.(*cloudformation.ElasticLoadBalancingV2Listener)
+				require.Equal(t, cloudformation.String("TCP"), listener.Protocol)
+				require.Equal(t, cloudformation.String("TCP"), tg.Protocol)
+			},
+		},
+		{
+			name: "NLBs should always use the TCP protocol for listeners and TGs - targetHTTPS should not impact it",
```
jonathanbeber, line 387: great, thanks!

aws/cf_template.go
```diff
@@ -325,7 +330,7 @@ func generateTemplate(spec *stackSpec) (string, error) {
 		HealthCheckIntervalSeconds: cloudformation.Ref(parameterTargetGroupHealthCheckIntervalParameter).Integer(),
 		HealthCheckPath:            cloudformation.Ref(parameterTargetGroupHealthCheckPathParameter).String(),
 		HealthCheckPort:            cloudformation.Ref(parameterTargetGroupHealthCheckPortParameter).String(),
-		HealthCheckProtocol:        cloudformation.String("HTTP"),
+		HealthCheckProtocol:        cloudformation.String(protocol),
```
jonathanbeber, line 333: I think that's what @mikkeloscar was referring to. We can do that and it will work for new load balancers, but old ones will fail to update because AWS does not support updating the healthcheck protocol of a running LB, maybe it's just for NLBs?

mikkeloscar: No, for ALBs there is not this limitation, there you can change the protocol. Just tested it :)

mikkeloscar:
:+1:

aws/cf_template.go
jhuntwork: I actually tested this in the real world already, and for my ALB, cloudformation just reconfigured it. It think it replaced the target group with a new one, but the load balancer and listener remained the same.

jonathanbeber: So it's a NLB ~bug~ feature

jonathanbeber:
:+1:

jonathanbeber (approved):

jonathanbeber:
Thanks, @jhuntwork, you should see an automated release soon.

aws/cf_template.go
```diff
@@ -112,11 +115,13 @@ func generateTemplate(spec *stackSpec) (string, error) {
...
 
-	protocol := "HTTP"
-	tlsProtocol := "HTTPS"
+	protocol := httpProtocol
+	tlsProtocol := httpsProtocol
 	if spec.loadbalancerType == LoadBalancerTypeNetwork {
 		protocol = "TCP"
 		tlsProtocol = "TLS"
+	} else if spec.targetHTTPS {
+		protocol = httpsProtocol
```
AlexanderYastrebov, line 124: This changes protocol of the targetGroup but also protocol of the _HTTP_ listener:
https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/92fa334e0d6332d34586aa226d2b2b00dc3453d0/aws/cf_template.go#L159

If I am not mistaken HTTP Listener protocol should be either "HTTP" or "TCP" in case of NLB

AlexanderYastrebov: I have created https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/434 if that makes sense

PR #412, opened 2021-03-30
:bug: Add possible HTTPS support on new stacks

jhuntwork:
I missed the different stackSpec setup for the CreateStack method, and
only added for UpdateStack. Side effect is that new stacks would always
still have a HTTP target group, but stack updates would change it to
HTTPS if specified.

Signed-off-by: Jeremy Huntwork <jhuntwork@lightcubesolutions.com>

### DIFF
--- a/aws/adapter.go
+++ b/aws/adapter.go
@@ -595,6 +595,7 @@ func (a *Adapter) CreateStack(certificateARNs []string, scheme, securityGroup, o
 			timeout:  a.healthCheckTimeout,
 		},
 		targetPort:                        a.targetPort,
+		targetHTTPS:                       a.targetHTTPS,
 		timeoutInMinutes:                  uint(a.creationTimeout.Minutes()),
 		stackTerminationProtection:        a.stackTerminationProtection,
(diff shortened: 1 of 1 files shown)


jhuntwork:
I wasn't sure how to test for this case. Any suggestions?

jonathanbeber:
Usually, this kind of setup we add to our e2e tests on [zalando-incubator/kubernetes-on-aws](https://github.com/zalando-incubator/kubernetes-on-aws), but indeed we should think about how to add at least some tests here. I will try to setup something soon.

jonathanbeber:
@mikkeloscar if you have a better ideas for this kind of integration tests I'm happy to hear it.

mikkeloscar:
:+1:

mikkeloscar:
@jonathanbeber I agree it's a bit tricky to test this as we really need an AWS environment to not mock the whole CF API (and thus not test much anyway).

szuecs:
@jonathanbeber @mikkeloscar maybe we should do golden files style tests, such that we have a set of cf stacks, maybe empty and input data („from“ kubernetes) as code configuration for table tests and check the stack output we would sent to Aws APIs, without mocking.

PR #424, opened 2021-07-26
Handle deleted certs by removing them from the LB stacks

mikkeloscar:
If a certificate defined on a stack is manually deleted from ACM the controller will end up in a situation where it keeps the certificate on the stack and continuously tries to update the stack which fails because of the non-existing certificate.

This PR aims to resolve this by checking if the certificate still exists when it builds an internal representation of the load balancer based on the stack, if it no longer exists it's not added and will be removed from the stack in the next update.

Additionally to this we also need to run the controller with a lower `--cert-polling-interval` to reduce the time where the controller thinks a deleted certificate exists. I don't change the default (`30m`) in this PR, but we may consider changing it after running with a lower value in Zalando for some time.

### DIFF
--- a/worker.go
+++ b/worker.go
@@ -22,19 +22,20 @@ import (
 )
 
 type loadBalancer struct {
-	ingresses        map[string][]*kubernetes.Ingress
-	scheme           string
-	stack            *aws.Stack
-	shared           bool
-	http2            bool
-	clusterLocal     bool
-	securityGroup    string
-	sslPolicy        string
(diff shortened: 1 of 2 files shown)


szuecs:
@mikkeloscar regarding `--cert-polling-interval`, I would say that a restart would trigger an immediate update, so this should be fine.

szuecs:
:+1:

jonathanbeber:
:+1:

Issue #428, opened 2021-09-01
use ALB in case ALB only feature is selected


szuecs:
Right now if your the controller with `--load-balancer-type=network`, NLBs are created as default. In case you run this controller with defaults or with `--load-balancer-type=application` for a while and want to change to NLBs as default, then the controller will switch all ingress/routegroup resources to NLBs, that do not explicitly use the `zalando.org/aws-load-balancer-type: alb` annotation.
This seems to be a bug, because in case someone decides to use ALB features like WAF (`zalando.org/aws-waf-web-acl-id`) or SG (`zalando.org/aws-load-balancer-security-group`), then the user obviously wants to have the enabled feature.

PR #430, opened 2021-09-01
fix: assume alb in case ALB feature is in use by a kubernetes resource

szuecs:
fix https://github.com/zalando-incubator/kube-ingress-aws-controller/issues/428


AlexanderYastrebov:
:+1:

mikkeloscar:
I think we also need to consider this inside the template generation logic because we do different things depending on `spec.loadbalancerType`. We have some tests for how the final cloudformation stack looks like which could be used to validate these cases.

szuecs:
Then I think the best is to move the code to when we create spec in the adapter

szuecs:
@mikkeloscar the problem with the CF stack generation is that they depend on

PR #431, opened 2021-09-01
Refactors kubernetes adapter `UpdateIngressLoadBalancer` method

AlexanderYastrebov:
Update of ingress or routegroup load balancer requires only namespace and name
parameters while hostname change check could be performed upfront.

The change removes unnecessary mapping to the internal structures `*ingress` and `*routegroup`.
This also simplifies "roundtrip" ingress mapping test into just "forward"
which enables testcases for empty annotations.


szuecs:
:+1:

PR #432, opened 2021-09-02
Default NLB fallbacks to ALB in case if incompatible feature requested

### LINKED ISSUE #428: use ALB in case ALB only feature is selected

AlexanderYastrebov:
Fixes #428 


kubernetes/adapter.go
```diff
@@ -217,7 +217,27 @@ func (a *Adapter) parseAnnotations(annotations map[string]string) *Ingress {
...
+	securityGroup, hasSG := annotations[ingressSecurityGroupAnnotation]
+	if !hasSG {
+		securityGroup = a.ingressDefaultSecurityGroup
+	}
+
+	wafWebAclId, hasWAF := annotations[ingressWAFWebACLIDAnnotation]
+
+	if (loadBalancerType == loadBalancerTypeNLB) && (hasSG || hasWAF) {
+		if hasLB {
+			log.Warn("Security Group and WAF are not supported by NLB (configured by annotation)")
```
szuecs, line 234: seems more Info level to me. The user decides that she wants to enforce NLB. It might be useful to have hasSG or hasWAF printed.

AlexanderYastrebov, line 234: Maybe this message should be enriched with ingress namespace+name

kubernetes/adapter.go
```diff
@@ -217,7 +217,27 @@ func (a *Adapter) parseAnnotations(annotations map[string]string) *Ingress {
...
+		} else {
+			log.Warn("Security Group and WAF are not supported by NLB (default), falling back to ALB")
```
szuecs, line 236: IMO that is Info or even better Debug, because this will be printed in too many cases.

kubernetes/adapter.go
AlexanderYastrebov: Also in this block we should maybe unset the config
```go
securityGroup = ""
wafWebAclId = ""
```

kubernetes/adapter_test.go
```diff
@@ -23,18 +23,19 @@ var (
 	testIPAddressTypeDualStack      = aws.IPAddressTypeDualstack
 	testIPAddressTypeDefault        = aws.IPAddressTypeIPV4
 	testLoadBalancerTypeIngress     = loadBalancerTypeALB
-	testLoadBalancerTypeAWS         = aws.LoadBalancerTypeApplication
```
AlexanderYastrebov, line 26: Inlined

szuecs (commented):
lgtm, but logging. I am quite picky on warning. My relatively strong opinion is that warning level is in most cases only for deprecation warnings.

szuecs:
:+1:

kubernetes/adapter.go
```diff
@@ -217,7 +211,29 @@ func (a *Adapter) parseAnnotations(annotations map[string]string) *Ingress {
...
+		securityGroup = a.ingressDefaultSecurityGroup
+	}
+
+	wafWebAclId, hasWAF := annotations[ingressWAFWebACLIDAnnotation]
+
+	if (loadBalancerType == loadBalancerTypeNLB) && (hasSG || hasWAF) {
+		log := log.WithFields(log.Fields{"type": typ, "ns": metadata.Namespace, "name": metadata.Name})
+
+		if hasLB {
+			log.Info("Security Group or WAF are not supported by NLB (configured by annotation)")
```
mikkeloscar, line 230: It seems to me like this should be an error and we should not create/update the ingress in this case but ignore it. Otherwise it's maybe confusing for the user that it was configured but not with the intended settings.

(ofc. we could also argue that this could be caught at admission/deployment, before the ingress resource is even submitted to the cluster).

szuecs: @mikkeloscar I have to disagree.
Imagine the situation is like this:
You have a WAF and default to ALB. Later you switch default to NLB, which will do nothing and stay at ALB.
Now you have an incident that can be mitigated by NLB. Hurry up and you set the lb annotation to specify "I WANT AN NLB!!!!". This is what I imagine in this case, so I think it is right to use NLB in this case.

szuecs: After discussions we agreed on to follow @mikkeloscar 's idea. Errors in configuration will be seen only in logs.

kubernetes/adapter_test.go
```diff
@@ -223,9 +227,180 @@ func TestNewIngressFromKube(tt *testing.T) {
...
+					LoadBalancer: ingressLoadBalancerStatus{
+						Ingress: []ingressLoadBalancer{
+							{Hostname: "bar"},
+						},
+					},
+				},
+			},
+		},
+		{
+			msg:                     "test explicitly configured NLB with security group does not fallback to ALB",
```
szuecs, line 331: should error now or not?

kubernetes/adapter_test.go
```diff
@@ -223,9 +227,180 @@ func TestNewIngressFromKube(tt *testing.T) {
...
+					LoadBalancer: ingressLoadBalancerStatus{
+						Ingress: []ingressLoadBalancer{
+							{Hostname: "bar"},
+						},
+					},
+				},
+			},
+		},
+		{
+			msg:                     "test explicitly configured NLB with WAF does not fallback to ALB",
```
szuecs, line 366: should error now or not?

szuecs:
I think we should also update readme.md for changes in minor version increase and increase minor in https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/master/VERSION

AlexanderYastrebov:
I have updated the PR to log error and ignore `(SG|WAF)&NLB` configuration

kubernetes/adapter_test.go
AlexanderYastrebov: fixed

szuecs:
:+1:

aryszka:
:+1:

PR #434, opened 2021-09-14
HTTPListener uses HTTP protocol when target is HTTPS

AlexanderYastrebov:
Followup on https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/411#discussion_r708516996

* Inlined well-known constants (HTTP, HTTPS, TCP, TLS)
* Added explicit load balancer type checks for both HTTP and HTTPS listeners
* Moved target group protocol and health check protocol selection close to target group construction

### DIFF
--- a/aws/cf_template.go
+++ b/aws/cf_template.go
@@ -24,9 +24,6 @@ const (
 	//
 	// [0]: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticloadbalancingv2-listenerrule.html#cfn-elasticloadbalancingv2-listenerrule-priority
 	internalTrafficDenyRulePriority int64 = 1
-
-	httpProtocol  = "HTTP"
-	httpsProtocol = "HTTPS"
 )
 
 func hashARNs(certARNs []string) []byte {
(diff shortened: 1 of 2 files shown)


AlexanderYastrebov:
FYI @jhuntwork

aws/cf_template.go
```diff
@@ -227,7 +244,7 @@ func generateTemplate(spec *stackSpec) (string, error) {
 		resourceName := fmt.Sprintf("HTTPSListenerCertificate%x", hashARNs(certificateARNs))
 		template.AddResource(resourceName, &cloudformation.ElasticLoadBalancingV2ListenerCertificate{
 			Certificates: &certificateList,
-			ListenerArn:  cloudformation.Ref(listenerName).String(),
+			ListenerArn:  cloudformation.Ref("HTTPSListener").String(),
```
szuecs, line 247: I don't like the repetition of the inline string, maybe we can have a const for this.
We create a resource and we reference it with the same string, so maybe a const would be the right choice.

szuecs:
:+1:

aws/cf_template.go
AlexanderYastrebov: well if we have
```go
const HTTPListener = "HTTPListener"
const HTTPSListener = "HTTPSListener"
```
it is as easy to mistype `HTTPListener vs HTTPSListener`  as `"HTTPListener"/"HTTPSListener"` but of course prevents `"HTTPXListener"`.

I have inlined
```go
httpProtocol  = "HTTP"
httpsProtocol = "HTTPS"
```
to remove the indirection - the `"HTTP"/"HTTPS"` are part of cloud formation DSL I do not see a reason why to put them into constant.

I am not strictly against constants here but I think we should tackle them systematically (needs thinking) and not only in a separate cases like this.

jhuntwork:
Nice. To be perhaps even clearer, NLBs don't exactly support HTTP either. It's either [TCP, UDP, TCP_UDP, or TLS](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/load-balancer-listeners.html#listener-configuration). "HTTP" in this case would presumably be TCP on port 80, but no actual HTTP protocol is being spoken there, it's just passing the TCP through. The target group config doesn't technically allow for HTTP either, your choices there are TCP or TLS, same idea. It's the registered targets that would actually speak the protocol, although you can configure HTTP or HTTPS health checks.

The TLS option should allow you to listen correctly on port 443 with a cert to offload the encryption, and then route the tcp on to the target group as HTTP. There are some options for the TLS ALPN extension that allows you to negotiate which HTTP protocols are accepted at that early point, but that seems to be the extent of it.

Anyway, I think the changes looks conceptually good, but it might just be good to be explicit about those terms in the documentation and flags.

aws/cf_template.go
szuecs: I have no strong opinion on that, so let's go with this

aws/cf_template.go
```diff
@@ -115,39 +112,53 @@ func generateTemplate(spec *stackSpec) (string, error) {
...
+						"HTTPListener",
+						internalTrafficDenyRulePriority,
+						spec.internalDomains,
+						spec.denyInternalDomainsResponse,
+					),
+				)
+			}
+		}
+	} else if spec.loadbalancerType == LoadBalancerTypeNetwork && spec.nlbHTTPEnabled {
+		template.AddResource("HTTPListener", &cloudformation.ElasticLoadBalancingV2Listener{
```
szuecs, line 161: maybe "TCPListener" instead of "HTTPListener"

aws/cf_template.go
```diff
@@ -183,35 +182,53 @@ func generateTemplate(spec *stackSpec) (string, error) {
...
+					generateDenyInternalTrafficRule(
+						"HTTPSListener",
+						internalTrafficDenyRulePriority,
+						spec.internalDomains,
+						spec.denyInternalDomainsResponse,
+					),
+				)
+			}
+		} else if spec.loadbalancerType == LoadBalancerTypeNetwork {
+			template.AddResource("HTTPSListener", &cloudformation.ElasticLoadBalancingV2Listener{
```
szuecs, line 215: "HTTPSListener" -> "TLSListener"

szuecs:
I marked the parts, that @jhuntwork referred to but for me it looks like the referencing would happen in one place later regardless of type ALB or NLB, so it would break, if we blindly execute what I commented inline.

AlexanderYastrebov:
@jhuntwork Thank you for having a look. 

Just to clarify, the change is not related to NLB but to ALB (see test change). For ALB we always have two listeners HTTP and HTTPS: 
- `HTTPListener` uses HTTP protocol and 80 port
- `HTTPSListener` uses HTTPS protocol and 443 port

With `target-https=true` before this change `HTTPListener` would be configured to use HTTP**S** protocol and 80 port (see https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/411#discussion_r708516996) which seems to be wrong because then we would have two listeners talking HTTPS on two ports 80 and 443.

In my understanding clients could both
- talk HTTPS to 443 port listener and listener would talk HTTPS to the target group
- talk HTTP to 80 port listener and listener would talk HTTPS to the target group

HTTP listener talking HTTPS to the target group feels a bit stange but it is consistent with HTTPSListener that talks HTTPS.
Whether we should have both HTTP and HTTPS listeners when target uses HTTPS is another story but here I aim to fix the above and restructure the code a bit for the upcoming changes.

Please correct me if I misunderstand something.

@szuecs 
Regarding listener names - HTTP rather means 80 port and HTTPS means 443; we could think of them as `Listener80` and `Listener443` where actual protocol depends on the loadbalancer type. I decided not to rename them here (and I am also not sure yet rename would not trigger massive LB recreation once deployed)

szuecs:
@AlexanderYastrebov NLB listener is not talking HTTPS nor HTTP, but TCP/TLS,UDP. That's the only thing I wanted to highlight as far as I understood @jhuntwork 's comment

I agree with you that the TG should be either HTTPS or HTTP based on the target and not based on the Listener.
I agree that renaming is not important and could harm more than being useful. That's why I :+1: :)

jhuntwork:
Sounds good, thanks for the clarification. Is there a container image that gets built with this PR I can test?

szuecs:
@jhuntwork yes the current should be: registry.opensource.zalan.do/teapot/kube-ingress-aws-controller-test:pr-434-1

jhuntwork:
I believe my comments about the NLB were due to a misreading of @AlexanderYastrebov's [comments in the related MR](https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/411#discussion_r708516996), sorry for that confusion. But perhaps it does illustrate the importance of good naming/documentation, especially for the two very different use cases between NLBs and ALBs. I understand though if renaming in the code currently presents another set of challenges that is too much to take on right now.

szuecs:
@AlexanderYastrebov then let's merge it and we test with our e2e test in kubernetes-on-aws.

AlexanderYastrebov:
:+1:

PR #435, opened 2021-09-16
Adds support for configurable HTTP target port

AlexanderYastrebov:
New `--alb-http-target-port` and `--nlb-http-target-port` flags allow to configure a second target
port distinct from `--target-port` that would be used by an HTTP listener.

In order to support two target ports controller creates two target
groups that use the same healthcheck configuration.


aws/cf.go
```diff
@@ -167,6 +176,13 @@ type denyResp struct {
 	body        string
 }
 
+func (spec *stackSpec) httpTargetPort() uint {
+	if spec.loadbalancerType == LoadBalancerTypeNetwork && spec.nlbHTTPEnabled && spec.nlbHTTPTargetPort != 0 {
+		return spec.nlbHTTPTargetPort
+	}
+	return spec.targetPort
+}
```
AlexanderYastrebov, line 184: This could be extended to support second target port for ALB in the similar way:
```go
...
} else if spec.loadbalancerType == LoadBalancerTypeApplication && spec.albHTTPTargetPort != 0 {
    return spec.albHTTPTargetPort
}
return spec.targetPort
```

AlexanderYastrebov: The idea is to detect of two ports are used via
```go
if port := spec.httpTargetPort(); port != spec.targetPort {
...
}
```

controller.go
```diff
@@ -155,6 +156,8 @@ func loadSettings() error {
 		Default("false").BoolVar(&nlbCrossZone)
 	kingpin.Flag("nlb-http-enabled", "Enable HTTP (port 80) for Network Load Balancers. By default this is disabled as NLB can't provide HTTP -> HTTPS redirect.").
 		Default("false").BoolVar(&nlbHTTPEnabled)
+	kingpin.Flag("nlb-http-target-port", "Sets the target port for NLB HTTP listener different from --target-port. Requires --nlb-http-enabled.").
+		UintVar(&nlbHTTPTargetPort)
```
AlexanderYastrebov, line 160: The new flag enables second target port only for NLB, for ALB both HTTP and HTTPS listeners continue to use single target and port. This could be extended in the same way for ALB by adding another `--alb-http-target-port` flag.

The alternative is to add generic `--http-target-port` flag that would enable second port for HTTP listener for both ALB and NLB.

szuecs: I think `--alb-http-target-port` makes sense to have, too. We can support both and are fine for any use case.

AlexanderYastrebov: I have added `--alb-http-target-port`

aws/cf.go
AlexanderYastrebov: I have added `--alb-http-target-port`

aws/cf.go
```diff
@@ -130,6 +138,8 @@ type stackSpec struct {
 	vpcID                             string
 	healthCheck                       *healthCheck
 	targetPort                        uint
+	albHTTPTargetPort                 uint
+	nlbHTTPTargetPort                 uint
```
AlexanderYastrebov, line 142: No need to have two fields as the stack can only have one LB type

aws/cf.go
```diff
@@ -109,7 +116,8 @@ const (
 	parameterTargetGroupHealthCheckPortParameter     = "TargetGroupHealthCheckPortParameter"
 	parameterTargetGroupHealthCheckIntervalParameter = "TargetGroupHealthCheckIntervalParameter"
 	parameterTargetGroupHealthCheckTimeoutParameter  = "TargetGroupHealthCheckTimeoutParameter"
-	parameterTargetTargetPortParameter               = "TargetGroupTargetPortParameter"
+	parameterTargetGroupTargetPortParameter          = "TargetGroupTargetPortParameter"
```
AlexanderYastrebov, line 119: Fixed name parameterTargetTargetPortParameter -> parameterTarget**Group**TargetPortParameter

aws/cf_test.go
```diff
@@ -58,6 +58,32 @@ func TestCreatingStack(t *testing.T) {
...
+			"stack with ALB http port",
+			stackSpec{
+				name:             "foo",
+				securityGroupID:  "bar",
+				vpcID:            "baz",
+				loadbalancerType: LoadBalancerTypeApplication,
+				httpTargetPort:   7777,
+			},
+			cfMockOutputs{createStack: R(mockCSOutput("fake-stack-id"), nil)},
+			"fake-stack-id",
```
AlexanderYastrebov, line 71: The whole test suite apparently does not test anything besides that mock output matches expected value. 
I've added test to not reduce the coverage but the whole test needs refactoring.

szuecs: I agree

szuecs:
:+1:

mikkeloscar:
:+1:

PR #436, opened 2021-09-28
Do not attach nonexistent target groups

AlexanderYastrebov:
When controller attempts to attach a batch of target groups that
contains nonexistent target group (could be deleted manually by mistake)
to the autoscaling group the attempt fails to pass validation which
prevents attachment of the other valid target groups from the batch.

Also:
* Improves error log messages
* Adds processChunked helper function
* Adds describeTargetGroups
* Fixes `detachARNs`/`validARNs` slices' capacity

### DIFF
--- a/aws/adapter.go
+++ b/aws/adapter.go
@@ -543,7 +543,7 @@ func (a *Adapter) UpdateTargetGroupsAndAutoScalingGroups(stacks []*Stack) {
 	for _, asg := range a.TargetedAutoScalingGroups {
 		// This call is idempotent and safe to execute every time
 		if err := updateTargetGroupsForAutoScalingGroup(a.autoscaling, a.elbv2, targetGroupARNs, asg.name, ownerTags); err != nil {
-			log.Errorf("failed to update target groups for autoscaling group %q: %v", asg.name, err)
+			log.Errorf("Failed to update target groups for autoscaling group %q: %v", asg.name, err)
 		}
 	}
 
@@ -552,7 +552,7 @@ func (a *Adapter) UpdateTargetGroupsAndAutoScalingGroups(stacks []*Stack) {
 	for _, asg := range nonTargetedASGs {
 		// This call is idempotent and safe to execute every time
 		if err := updateTargetGroupsForAutoScalingGroup(a.autoscaling, a.elbv2, nil, asg.name, ownerTags); err != nil {
-			log.Errorf("failed to update target groups for non-targeted autoscaling group %q: %v", asg.name, err)
+			log.Errorf("Failed to update target groups for non-targeted autoscaling group %q: %v", asg.name, err)
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


aws/asg.go
```diff
@@ -213,67 +206,56 @@ func updateTargetGroupsForAutoScalingGroup(svc autoscalingiface.AutoScalingAPI,
...
-	}
+		return err
+	})
+	return descs, err
+}
 
-	return descs, nil
+func describeTargetGroups(elbv2svc elbv2iface.ELBV2API) (map[string]struct{}, error) {
+	targetGroups := make(map[string]struct{})
+	err := elbv2svc.DescribeTargetGroupsPagesWithContext(context.TODO(), &elbv2.DescribeTargetGroupsInput{}, func(resp *elbv2.DescribeTargetGroupsOutput, lastPage bool) bool {
```
AlexanderYastrebov, line 252: BTW I am not sure why do we use paged API with context as neither is utilized. There is a simple https://docs.aws.amazon.com/sdk-for-go/api/service/elbv2/#ELBV2.DescribeTargetGroups

jonathanbeber: I might be interpreting it wrong, but IMHO we do use the paging, we are returning always `true` in the paging function, which means we will always get all the content but in multiple chunks. The context is indeed not being used, I believe we planned to use it in the future for easier cancelling (!?). Dunno.

aws/asg.go
```diff
@@ -322,31 +304,47 @@ func hasTagsASG(tags []*autoscaling.TagDescription, expectedTags map[string]stri
 	return true
 }
 
+func attachTargetGroupsToAutoScalingGroup(svc autoscalingiface.AutoScalingAPI, targetGroupARNs []string, autoScalingGroupName string) error {
+	// limit target group updates to 10 at a time since this is the limit
+	// in AWS API.
+	return processChunked(targetGroupARNs, 10, func(chunk []string) error {
```
jonathanbeber, line 310: nit :nail_care: since this is the "limit" from AWS I think a constant would fit us well.

aws/asg.go
```diff
@@ -213,67 +206,56 @@ func updateTargetGroupsForAutoScalingGroup(svc autoscalingiface.AutoScalingAPI,
...
-			tgParams := &elbv2.DescribeTagsInput{
-				ResourceArns: aws.StringSlice(arnsChunked),
-			}
-
-			tgResp, err := svc.DescribeTags(tgParams)
-			if err != nil {
-				return nil, err
-			}
-
+	err := processChunked(arns, 20, func(chunk []string) error {
```
jonathanbeber, line 238: I wonder why it's 20 here :confused:

jonathanbeber:
:+1:

aws/asg.go
AlexanderYastrebov: I mean we do not use the feature to stop paging early as we always return true so if I do not miss anything the outcome is exactly as using `ELBV2.DescribeTargetGroups`

aws/asg.go
AlexanderYastrebov: `10` is a constant ;)

AlexanderYastrebov: I have added a link to the docs that specify the limit

aws/asg.go
AlexanderYastrebov: I have added a link to the docs that specify the limit

aws/asg.go
mikkeloscar: Normally such describe calls in the AWS APIs only list the first `n` (usually around 100) items. From the docs it looks like this one should list all of them, but I would not trust the docs to be right about this as it's usually not the case.

jonathanbeber: yeah, my initial understanding is that `DescribeTargetGroups` would return just the first page.

jonathanbeber:
:+1:

aws/adapter.go
```diff
@@ -552,21 +552,21 @@ func (a *Adapter) UpdateTargetGroupsAndAutoScalingGroups(stacks []*Stack) {
 	for _, asg := range nonTargetedASGs {
 		// This call is idempotent and safe to execute every time
 		if err := updateTargetGroupsForAutoScalingGroup(a.autoscaling, a.elbv2, nil, asg.name, ownerTags); err != nil {
-			log.Errorf("UpdateTargetGroupsAndAutoScalingGroups() failed to attach target groups to ASG '%s': %v", asg.name, err)
+			log.Errorf("failed to update target groups for non-targeted autoscaling group %q: %v", asg.name, err)
```
szuecs, line 555: Log messages should start with upper case

aws/asg.go
szuecs: As far as I remember it was exactly the problem we had found: list only the first `n` items, so paging was implemented

szuecs:
:+1:

aws/adapter.go
AlexanderYastrebov: Fixed

szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #440, opened 2021-10-06
Support for the latest FS policy on application load balancers

jbilliau-rcd:
There is a newer FS policy available on ALB's that this controller doesn't support. This would add it.


szuecs:
please rebase or amend with sign-off, thanks (DCO)

PR #441, opened 2021-10-06
Add newer FS policy support

jbilliau-rcd:
There is a newer FS policy available on ALB's that this controller doesn't support. This would add it.


szuecs:
ok then do the following: pull --rebase the current master, do the change and commit with `-s`, then push and open PR.

jbilliau-rcd:
Ok i'll give it a shot

PR #453, opened 2021-11-18
adding config option for ALB & NLB healthy threshold count

universam1:
Healthy threshold: The number of consecutive health checks successes required before considering an unhealthy target healthy

This configuration is required since the default of "5" time the interval, which is minimum of 6s, sums up to 30s that is too long delay in our configurations.

Signed-off-by: Samuel Lang <gh@lang-sam.de>

### DIFF
--- a/aws/cf_template.go
+++ b/aws/cf_template.go
@@ -446,9 +446,12 @@ func generateDenyInternalTrafficRule(listenerName string, rulePriority int64, in
 func newTargetGroup(spec *stackSpec, targetPortParameter string) *cloudformation.ElasticLoadBalancingV2TargetGroup {
 	protocol := "HTTP"
 	healthCheckProtocol := "HTTP"
+	healthyThresholdCount, unhealthyThresholdCount := spec.albHealthyThresholdCount, spec.albUnhealthyThresholdCount
 	if spec.loadbalancerType == LoadBalancerTypeNetwork {
 		protocol = "TCP"
 		healthCheckProtocol = "HTTP"
+		// For NLBs the healthy and unhealthy threshold count value must be equal
+		healthyThresholdCount, unhealthyThresholdCount = spec.nlbHealthyThresholdCount, spec.nlbHealthyThresholdCount
 	} else if spec.targetHTTPS {
 		protocol = "HTTPS"
 		healthCheckProtocol = "HTTPS"
@@ -465,6 +468,8 @@ func newTargetGroup(spec *stackSpec, targetPortParameter string) *cloudformation
 		HealthCheckPath:            cloudformation.Ref(parameterTargetGroupHealthCheckPathParameter).String(),
 		HealthCheckPort:            cloudformation.Ref(parameterTargetGroupHealthCheckPortParameter).String(),
 		HealthCheckProtocol:        cloudformation.String(healthCheckProtocol),
+		HealthyThresholdCount:      cloudformation.Integer(int64(healthyThresholdCount)),
+		UnhealthyThresholdCount:    cloudformation.Integer(int64(unhealthyThresholdCount)),
 		Port:                       cloudformation.Ref(targetPortParameter).Integer(),
 		Protocol:                   cloudformation.String(protocol),
 		VPCID:                      cloudformation.Ref(parameterTargetGroupVPCIDParameter).String(),
--- a/controller.go
+++ b/controller.go
@@ -43,6 +43,9 @@ var (
 	healthCheckPort               uint
 	healthCheckInterval           time.Duration
 	healthCheckTimeout            time.Duration
+	albHealthyThresholdCount      uint
+	albUnhealthyThresholdCount    uint
+	nlbHealthyThresholdCount      uint
 	targetPort                    uint
 	albHTTPTargetPort             uint
 	nlbHTTPTargetPort             uint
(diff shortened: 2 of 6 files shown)


universam1:
@szuecs please review, thanks

dhohengassner:
@universam1 nice addition - thanks for adding :+1:

w-reichert:
Good idea, I like that addition, too. 👍

universam1:
@szuecs Not sure what to do about the coverage decrease.

Tested this change in our clusters, works as expected

szuecs:
@universam1 it's no an issue if coverage drops <1%. It's more a hint we should make it better in the future.

szuecs:
:+1:

controller.go
```diff
@@ -221,6 +222,8 @@ func loadSettings() error {
 		Default(aws.DefaultHealthCheckInterval.String()).DurationVar(&healthCheckInterval)
 	kingpin.Flag("health-check-timeout", "sets the health check timeout for the created target groups. The flag accepts a value acceptable to time.ParseDuration").
 		Default(aws.DefaultHealthCheckTimeout.String()).DurationVar(&healthCheckTimeout)
+	kingpin.Flag("healthy-threshold-count", "The number of consecutive health checks successes required before considering an unhealthy target healthy").
+		Default(strconv.FormatUint(aws.DefaultHealthyThresholdCount, 10)).UintVar(&healthyThresholdCount)
```
AlexanderYastrebov, line 226: We need to think more about defaults:
* it is 5 for ALB https://docs.aws.amazon.com/elasticloadbalancing/latest/application/target-group-health-checks.html
* it is 3 for NLB https://docs.aws.amazon.com/elasticloadbalancing/latest/network/target-group-health-checks.html

One way is to have two flags `alb-healthy-threshold-count` and `nlb-healthy-threshold-count` (like we have `alb-http-target-port` and `nlb-http-target-port`).

jhuntwork: Nice, good catch. I recall encountering that when doing some other testing, too.

AlexanderYastrebov: There is also `UnhealthyThresholdCount` which is default 2 for ALB but must be equal to `HealthyThresholdCount` for NLB (not sure if it breaks when not specified). I guess we can leave it out of scope of this change until needed.

AlexanderYastrebov: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticloadbalancingv2-targetgroup.html#cfn-elasticloadbalancingv2-targetgroup-healthythresholdcount

universam1: Thanks @AlexanderYastrebov for that research, I missed that. I can add that prefix. Should we leave the NLB part out for now, as not required?

universam1: @AlexanderYastrebov scope is now reduced to ALBs only, the pattern is set so it can be extended easily if needed. Or do you like to do the full setup of all conditions now?

AlexanderYastrebov: > the pattern is set so it can be extended easily if needed

True, I do not think the full set of flags is necessary atm. 
@szuecs WDYT?

aws/cf_template.go
```diff
@@ -473,6 +473,7 @@ func newTargetGroup(spec *stackSpec, targetPortParameter string) *cloudformation
 	// custom target group healthcheck only supported when the target group protocol is != TCP
 	if protocol != "TCP" {
 		targetGroup.HealthCheckTimeoutSeconds = cloudformation.Ref(parameterTargetGroupHealthCheckTimeoutParameter).Integer()
+		targetGroup.HealthyThresholdCount = cloudformation.Integer(int64(spec.albHealthyThresholdCount))
```
AlexanderYastrebov, line 476: For some reason all other variables are referenced as template parameters. I am not sure yet why but I think we should follow the suit here.

AlexanderYastrebov: @szuecs @mikkeloscar Do you remember why do we use parameter refs elsewhere?

AlexanderYastrebov: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html

controller.go
szuecs: We can go with ALB only, but on the other hand I would recommend using NLBs anyways, so why not include it.
If you only need limited scope feel free to reduce the scope to ALB.

aws/cf_template.go
universam1: we discourage using parameter when the template is rendered by a dynamic language already like this project is, as it has a couple of downsides (such as no CFN diff), but I'm happy to adjust to comply whatever is desired

controller.go
universam1: Not a deal, I've added the support for ALBs and NLBs including their conditions. 👍🏻 
Hope that's it, let me know if there is anything else you like to have adjusted

controller.go
```diff
@@ -221,6 +224,12 @@ func loadSettings() error {
...
 	kingpin.Flag("health-check-timeout", "sets the health check timeout for the created target groups. The flag accepts a value acceptable to time.ParseDuration").
 		Default(aws.DefaultHealthCheckTimeout.String()).DurationVar(&healthCheckTimeout)
+	kingpin.Flag("alb-healthy-threshold-count", "The number of consecutive successful health checks required before considering an unhealthy target healthy. The range is 2–10. The default is 5. (ALB only)").
+		Default(strconv.FormatUint(aws.DefaultAlbHealthyThresholdCount, 10)).UintVar(&albHealthyThresholdCount)
+	kingpin.Flag("alb-unhealthy-threshold-count", "The number of consecutive failed health checks required before considering a target unhealthy. The range is 2–10. The default is 2. (ALB only)").
+		Default(strconv.FormatUint(aws.DefaultAlbUnhealthyThresholdCount, 10)).UintVar(&albUnhealthyThresholdCount)
+	kingpin.Flag("nlb-healthy-threshold-count", "The number of consecutive successful or failed health checks required before considering an target healthy or unhealthy. The range is 2 to 10. The default is 3. (NLB only)").
+		Default(strconv.FormatUint(aws.DefaultNlbHealthyThresholdCount, 10)).UintVar(&nlbHealthyThresholdCount)
```
AlexanderYastrebov, line 232: I think the default value is rendered in the help message:
```
  --alb-healthy-threshold-count=5  
                                 The number of consecutive successful health
                                 checks required before considering an unhealthy
                                 target healthy. The range is 2–10. The default
                                 is 5. (ALB only)
  --alb-unhealthy-threshold-count=2  
                                 The number of consecutive failed health checks
                                 required before considering a target unhealthy.
                                 The range is 2–10. The default is 2. (ALB only)
  --nlb-healthy-threshold-count=3  
                                 The number of consecutive successful or failed
                                 health checks required before considering an
                                 target healthy or unhealthy. The range is 2 to
                                 10. The default is 3. (NLB only)
```
so lets remove `The default is ...`.

For `--nlb-healthy-threshold-count` it should be `considering _a_ target` and I think we should use the same format for the range as for ALB: `The range is 2–10` for consistency (I get that it is different in the AWS docs).

controller.go
AlexanderYastrebov: :+1: Thank you

aws/cf_template.go
szuecs: I have no opinion on that.

AlexanderYastrebov: I think we should use parameters for consistency but I have opened an issue to reconsider https://github.com/zalando-incubator/kube-ingress-aws-controller/issues/454

AlexanderYastrebov: OTOH I see we are not using parameters for everything so maybe we should not bother introducing more now and it is ok as it is.

controller.go
universam1: Done, thanks for the review

AlexanderYastrebov: Thank you. `_an_ target` is still there though.

universam1: thanks, sorry

szuecs:
:+1:

Issue #391, opened 2021-01-14
HTTP/2 not enabled for target groups


universam1:
HTTP/2 is enabled per default for the load balancer, but for the target group!

According to this announcement and the AWS console the target groups do support HTTP/2 as well.

https://aws.amazon.com/blogs/aws/new-application-load-balancer-support-for-end-to-end-http-2-and-grpc/

However, the target groups are always configured to forward to connect via HTTP/1.1

This might have a performance impact

szuecs:
@universam1 normally it should not have an impact because of connection pools. Maybe for frontends it would benefit to do full H2 support.
Do you want to work on an option to enable H2 backend? 
I don't want to have this as default set, but the feature would be great to have!

universam1:
pending on Cloudformation support: https://github.com/aws-cloudformation/aws-cloudformation-coverage-roadmap/issues/713

universam1:
@szuecs Cloudformation support is [present](https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/713#issuecomment-778044121) in meantime, but the underlying [CFN lib fork](github.com/mweagle/go-cloudformation) seems to be unmaintained. 

Any idea @mikkeloscar how to continue here, guess that's a project relevant question?

szuecs:
I see. I think we have 2 possible ways:
- fork the dependency project 
- internal it

PR #458, opened 2021-12-09
Application Load Balancer Support for End-to-End HTTP/2

universam1:
ref: https://aws.amazon.com/de/blogs/aws/new-application-load-balancer-support-for-end-to-end-http-2-and-grpc/

Currently HTTP/2 is only half way implemented, that is the ALB can be configured for HTTP/2 but the TargetGroup is not configured.
This is probably due to the AWS delay of supporting that protocal version for the TargetGroup which is available in meantime.

There is however an update missing for `mweagle/go-cloudformation` which is deprecated, thus a fork has been made that includes this Cloudformation feature: https://github.com/o11n/go-cloudformation/commit/b975e65fc6d53c789602ceb28de10974ec69f87e

fixes https://github.com/zalando-incubator/kube-ingress-aws-controller/issues/391


universam1:
@szuecs I assume Zalando would be more in favor to fork `mweagle/go-cloudformation` themselves and apply the mentioned patch there?

aws/cf_template.go
```diff
@@ -478,6 +478,11 @@ func newTargetGroup(spec *stackSpec, targetPortParameter string) *cloudformation
 	// custom target group healthcheck only supported when the target group protocol is != TCP
 	if protocol != "TCP" {
 		targetGroup.HealthCheckTimeoutSeconds = cloudformation.Ref(parameterTargetGroupHealthCheckTimeoutParameter).Integer()
+		if spec.http2 {
+			targetGroup.ProtocolVersion = cloudformation.String("HTTP2")
+		} else {
+			targetGroup.ProtocolVersion = cloudformation.String("HTTP1")
+		}
```
AlexanderYastrebov, line 485: `spec.http2` is true by default. If I get it right this change would enable HTTP/2 on the target group and according to this table https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-target-groups.html#target-group-protocol-version



Request protocol | Protocol version | Result
-- | -- | --
HTTP/1.1 | HTTP/2 | Error
HTTP/2 | HTTP/2 | Success

this change
* is not backward-compatible as default protocol version is HTTP/1.1
* would not work if target supports only HTTP/1.1 
* will disable HTTP/1.1 incoming requests

If that is correct then we may probably need to have a new flag and/or annotation to steer Target Group protocol version independently and backward-compatible.

Besides that HTTP/2 is only supported in conjunction with `HTTPS` protocol so this kind of config should be around here
https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/5c661370ed2aa4cf2a0f4164f5a9114c80fe2c84/aws/cf_template.go#L455-L458

universam1: This is true, the http/2 TG protocol is a breaking change for http/1.1 clients. Looking for a procedure how to deal with while supporting end-to-end, found this statement: https://stackoverflow.com/a/66800142

Probably going to open a case at AWS

universam1:
> > I assume Zalando would be more inclined to fork mweagle/go-cloudformation themselves and apply the mentioned patch there?
> 
> The https://github.com/mweagle/go-cloudformation#this-repo-is-no-longer-supported suggests to move to https://github.com/awslabs/goformation which is most likely will be a large undertaking

This PR is naive mainly to trigger a decision where to go with the stale dependency

aws/cf_template.go
universam1: well known issue: https://forums.aws.amazon.com/message.jspa?messageID=985494

szuecs:
In the long run we should migrate to https://github.com/awslabs/goformation.
@universam1 do you want to do `ALB -> grpc-application` or do you want to pass this through skipper? 
For the latter I think there is some more work required for the former: why don't you use NLB instead?

aws/cf_template.go
szuecs: It also shows the complexity that everything is based on the client. I am thinking for a while to support h2c in skipper, but the current implementations have some critical bugs and if we would be able to serve it then I also do not know if grpc backends can listen on handlers without TLS....

universam1:
> In the long run we should migrate to https://github.com/awslabs/goformation. In general I don't mind to use a fork of an unmaintained library if this buys us some time to migrate to a more supported or better version. @universam1 do you want to do `ALB -> grpc-application` or do you want to pass this through skipper? For the latter I think there is some more work required for the former: why don't you use NLB instead?

Primary goal is to support http/2 pass-thru or end-to-end. Currently that is cross converted loosing the main benefits, see https://stackoverflow.com/a/36057190

@szuecs do you suggest replacing ALBs with NLBs for that use case? Not much experience with that type of LB, sounds counter intuitive using an OSI 4 LB for OSI 7 application

aws/cf_template.go
universam1: @AlexanderYastrebov interestingly the protocol H2 can be set via AWS console also for non-https target groups, also I found that documented in the TF provider https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lb_target_group#protocol_version

Nevertheless I implemented as requested

universam1:
@AlexanderYastrebov @szuecs despite of the AWS ALB compatibility limitations this PR is tested and it does work as expected. Added a feature toggle in order not to break compatibility.
Let me know if there is anything else, thanks

aws/cf_template_test.go
```diff
@@ -614,6 +619,38 @@ func TestGenerateTemplate(t *testing.T) {
 				require.NotEqual(t, cloudformation.Integer(3), tg.UnhealthyThresholdCount)
 			},
 		},
+		{
+			name: "target port http2 when https listener and configured",
```
dhohengassner, line 623: @universam1 should be there a test case for GRPC as well?

dhohengassner:
@universam1 thanks for implementing - good to have these additional options ready :+1:

aws/cf_template_test.go
szuecs: gRPC is using H2, so I guess this test is good enough, no?

szuecs:
@universam1 I just wanted to understand if possible options were thought about (NLB vs ALB with grpc TG). In general the less the load balancer has to do the better for latency. NLB would not break grpc I think, so that's why I asked. I have nothing against the PR itself and have no suggestions besides to answer other open comments.
Maybe @AlexanderYastrebov has more input.

universam1:
> as to do the better

universam1:
@AlexanderYastrebov  @szuecs  Sorry accidentally closed this PR by force pushing the commit, reopened here #460

szuecs:
no worries

PR #460, opened 2021-12-10
Application Load Balancer Support for End-to-End HTTP/2

universam1:
ref: https://aws.amazon.com/de/blogs/aws/new-application-load-balancer-support-for-end-to-end-http-2-and-grpc/

Currently HTTP/2 is only half way implemented, that is the ALB can be configured for HTTP/2 but the TargetGroup is not configured.
This is probably due to the AWS delay of supporting that protocal version for the TargetGroup which is available in meantime.

There is however an update missing for `mweagle/go-cloudformation` which is deprecated, thus a fork has been made that includes this Cloudformation feature: https://github.com/o11n/go-cloudformation/commit/b975e65fc6d53c789602ceb28de10974ec69f87e

fixes https://github.com/zalando-incubator/kube-ingress-aws-controller/issues/391


aws/cf_template_test.go
```diff
@@ -614,6 +619,38 @@ func TestGenerateTemplate(t *testing.T) {
 				require.NotEqual(t, cloudformation.Integer(3), tg.UnhealthyThresholdCount)
 			},
 		},
+		{
+			name: "target port http2 when https listener and configured",
```
szuecs, line 623: there was a comment by @dhohengassner  about adding a test with GRPC TG, which we miss in case you add the flag that supports GRPC in https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/460/files#diff-cc5d2cdea66aa1021f86124b6b59590af4179f66c5da18e04c43ae82c0b6bfbdR284

aws/cf_template.go
```diff
@@ -447,14 +447,17 @@ func newTargetGroup(spec *stackSpec, targetPortParameter string) *cloudformation
 	protocol := "HTTP"
 	healthCheckProtocol := "HTTP"
 	healthyThresholdCount, unhealthyThresholdCount := spec.albHealthyThresholdCount, spec.albUnhealthyThresholdCount
+	protocolVersion := cloudformation.String("HTTP1")
```
AlexanderYastrebov, line 450: Maybe we should keep it `nil` to be backward compatible?

controller.go
```diff
@@ -321,6 +324,10 @@ func loadSettings() error {
 		return fmt.Errorf("invalid max number of certificates per ALB: %d. AWS does not allow more than %d", maxCertsPerALB, aws.DefaultMaxCertsPerALB)
 	}
 
+	if !strings.Contains("HTTP1 HTTP2 GRPC", targetGroupProtocolVersion) {
+		return fmt.Errorf("invalid target group protocol version: %s. must be one of HTTP1, HTTP2, GRPC", targetGroupProtocolVersion)
+	}
+
```
AlexanderYastrebov, line 330: We need to decide also if we would like to support ingress annotation in this scope to enable this individually.
The default value is defined as `"HTTP1"` but before this change we did not set it so I am wondering if changing empty to `"HTTP1"` will cause any issues (e.g. trigger loadbalancer recreation).

Lets check if we can use `kingpin.Enum`. If no it would be still better to check three possible cases instead of this clever trick which e.g. allows `"HTTP1 HTTP2"` or `"TTP"`

controller.go
```diff
@@ -280,6 +281,8 @@ func loadSettings() error {
 		Default("text/plain").StringVar(&denyInternalRespContentType)
 	kingpin.Flag("deny-internal-domains-response-status-code", "Defines the response status code for a request identified as to an internal domain when -deny-internal-domains is set.").
 		Default("401").IntVar(&denyInternalRespStatusCode)
+	kingpin.Flag("target-group-protocol-version", "Defines the target group protocol version for ALB. Specify GRPC to send requests to targets using gRPC. Specify HTTP2 to send requests to targets using HTTP/2. The default is HTTP1, which sends requests to targets using HTTP/1.1. Note the compatibility chart in https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-target-groups.html#target-group-protocol-version.").
```
AlexanderYastrebov, line 284: Lets see if `kingpin.Enum` could be used to print possible values.
I assume users know what they configure so lets maybe just reference `See https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-target-groups.html#target-group-protocol-version` without describing the behaviors.

AlexanderYastrebov: `Default("HTTP1")` prints default value so there is no need to mention it in the help message.

controller.go
AlexanderYastrebov: /cc @szuecs

universam1: The update behavior of TG is recreating, which is causing a new resource to be created first and the replacing the link. But I'm fine to change accordingly.

universam1: I'm not sure about annotation support, since this change is invasive and potentially breaking HTTP1.1 clients, why I would see this as a central administrative decision

aws/cf_template_test.go
universam1: done

aws/cf_template.go
universam1: Could be problematic, consider the case going back from H2 to H1 where CF might not detect that config change due to the parameter being nil. I think the default value will not become mandatory again!?

controller.go
szuecs: Enum sounds like a good idea to me.
In general it should be the administrator that decides about the default via "flag" and developers can opt-in a different choice via annotation to override the global decision.

jbilliau-rcd:
Is this still being worked on? Having HTTP2 support down through the TG would be nice. RIght now, the traffic downgrades to HTTP1 which doesnt let us take advantage of all of the benefits of HTTP2.

szuecs:
We wait first for the test infrastructure changes before we add new features. Right now the colleague is in vacation. I hope next week we can finish the reviews and go on by merging it.

universam1:
> Cannot start a pipeline due to:
> 
> ```
> No accountable user for this pipeline: no Zalando employee associated to this GitHub username
> ```
> 
> Click on **pipeline** status check Details link below for more information.

@szuecs rebased the PR, let me know if anything else to do. Thanks

Issue #298, opened 2019-12-23
Support for target-type IP


rifelpet:
For CNI providers that provide VPC IPs to pods, ALBs and NLBs can target the pod IPs directly rather than instance IDs. This bypasses kube-proxy and can avoid cross-AZ traffic by instances that do not run one of the service's pods.

https://docs.aws.amazon.com/elasticloadbalancing/latest/network/load-balancer-target-groups.html#target-type

I think it would be great if kube-ingress-aws-controller could support targeting the IPs behind a Service (its Endpoints). An additional `target-type` annotation could be added to support this. The controller would need to watch Endpoints in order to keep target groups up to date when the ready pods behind a Service change. We're open to contributing this functionality if thats acceptable.

Thoughts?

szuecs:
@rifelpet from my side it’s fine to implement that, if you opt-in via cli flag to enable endpoint lookups and annotations to opt-in via ingress. I am happy to read your pull request.

On the other hand no one should use nodeport service and if you use skipper as hostnetwork target you get the same (no service but endpoints), of course you have an additional hop, which might be bad for some workloads with very tight latency demands.

rverma-jm:
+1

universam1:
Good to see there is a discussion started already. Made a test setup to POC with Skipper **not** using `hostPort` but  regular `containerPort` only, accessing via the VPC IP of these pods.

Basically, configure the ALB to use a target-group of `IP` instead of `Instance`, which are pointing to the Skipper pods IPs directly, and it works just fine!
That means, using AWS CNI it is possible to untie Skipper from hostPort!

advantages:
* Skipper is independent of nodes, it can scale freely horizontally in both directions
* no unhealthy endpoints on the `target group` when Skipper is run as deployment with replicas < nodes
* no extra hop introduced, compared to a "Service"
* PDB or scalingPolicies are better supported, because there can be more than one Skipper per node simplifying the rolling update

disadvantage:
* works only with AWS CNI (?)

Would that implementation make sense?

thoughts @szuecs

szuecs:
@universam1 sounds like a great step forward. If you can create an option such that you can enable this behavior as opt-in, I would be happy to review  the pr. 😊

PR #461, opened 2021-12-17
✨ AWS CNI Mode

universam1:
# new operational mode of Direct Ingress Pod access

implements #298 Support for target-type IP

This adds a new operational mode of adding the **Ingress Pods as Targets in the Target Groups directly**, instead of the current operation mode where the Ingress pods are accessed through a **HostPort**.

The core idea is based on the fact that standard AWS EKS cluster running AWS VPC CNI have their pods as first class members in the VPCs. Their IPs are directly reachable from the ALB/NLB target groups like the nodes are too, which means there is no necessity for the HostPort indirection.

There are several drivers and advantages accessing the pod directly vs a HostPort:

## HostNetwork and HostPort conflict with best practices

ref: https://kubernetes.io/docs/concepts/configuration/overview/#services

## Delayed, eventual consistency of AutoscalingGroup to ALB/NLB target group updates

This has been the biggest trouble in operations, that the update of node members in target groups is slower than the nodes are physically replaced, which ends up in a black hole of **no Ingresses available at a time**. We are facing regularly downtimes especially when spot interruptions or ASG node rollings happen that the ALB/NLB takes up to 2 minutes to reflect the group change. For smaller clusters this leads to no Skipper instance being registered hence no target available to forward traffic to.
With this new mode the registration happens independently of ASGs and instantly, the scheduling of pods up to be serving traffic from ALB takes less than 10 seconds!

## Independent scaling from nodes

With HostPort there is an eventual dependency on available nodes to scale the Ingress.
Plus the Ingress pod cannot be replaced in place but requires a termination first and then rescheduling. For a certain time, which can be more than a minute, this node is offline as an Ingress.
With this mode the (host networking ?) and HostPort is obsolete, which allows node independent scaling of Skipper pods! Skipper becomes a regular deployment and its ReplicaSet can be independent on the cluster size, which simplifies operations especially for smaller clusters. We are using a custom HPA metric attached to node group size to counteract this deployment / daemonset hybrid combination, which is obsolete now!

## Synced de/registering and instantaneous response

Core idea is the event based registration to Kubernetes using pod `Informer` that receives immediate notifications about pod changes, which allow almost zero delayed updates on the target groups.

The registration happens as soon as the pod received an IP from AWS VPC. Hence the readiness probe of the ALB/NLB starts to monitor already during scheduling of the pod, serving the earliest possible. Tests in lab show pods serving ALB traffic well under 10s from scheduling!

Deregistration happens bound to the kubernetes event. That means the LB is now in sync with the cluster and will stop sending traffic before the pod is actually terminated. This implement save deregistering without traffic loss. Tests in lab show even under aggressive deployment scalings there are no packet losses measured!

## TG without unhealthy targets

Since the IP based TGs are managed now by this controller, they represent pods and thus all of them are shown healthy, otherwise cleaned up by this controller.

# Implementation details:

* client-go Informer: This high level functions are providing a convenient access to event registrations of kubernetes. Since the event registration is the key of fast response and efficient compared to high rate polling, using this standard factory methods seems standing to reason.

## Tests

*  successful transistion of TG from type Instance to type IP vice versa
* the controller registers pods that are discovered
* the controller deregisters pods that are "terminating" status
* the controller recovers desired state if manual intervention on the TG happened by "resyncing"
* it removes pods that are killed or dead

### HostNetwork and HostPort combinations tested:

| access mode | HostNetwork | HostPort | Result |                      Notes                       |
| :---------: | :---------: | :------: | :----: | :----------------------------------------------: |
| `Instance`  |   `true`    |  `true`  |  `ok`  |                    status quo                    |
|    `IP`     |   `true`    |  `true`  |  `ok`  |       PodIP == HostIP --> limited benefits       |
|    `IP`     |   `false`   |  `true`  |  `ok`  | PodIP != HostIP --> limited scaling, node bounde |
|    `IP`     |   `true`    | `false`  | `N/A`  |            impossible, HN implies HP             |
|    `IP`     |   `false`   | `false`  |  `OK`  |            goal achieved: free scaling and HA             |

Example logs:
```
time="2021-12-17T15:36:36Z" level=info msg="Deleted Pod skipper-ingress-575548f66-k99vs IP:172.16.49.249" 
time="2021-12-17T15:36:36Z" level=info msg="Deregistering CNI targets: [172.16.49.249]" 
time="2021-12-17T15:36:37Z" level=info msg="New Pod skipper-ingress-575548f66-qff2q IP:172.16.60.193"
time="2021-12-17T15:36:40Z" level=info msg="Registering CNI targets: [172.16.60.193]"
```

## misc

* extended the service account with required RBAC permissions to watch/list pods
* added example of Skipper without a HostPort and HostNetwork


universam1:
fixes #298 

cc @szuecs

szuecs:
@universam1 thanks, sounds like a great idea for aws-vpc cni.
Does the change also work with plain hostnetwork without nodeport? 
We don't use nodeport service. see also https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/skipper/deployment.yaml#L47-L56

One drawback is that we need informers and because of that client-go WATCH (internal to informers). We tried to omit client-go, because of https://github.com/kubernetes/client-go/issues/374#issuecomment-413842964 https://github.com/zalando-incubator/stackset-controller/issues/263 , but since https://github.com/kubernetes/kubernetes/pull/95981 was done maybe it's also fine to use client-go.

universam1:
Thanks @szuecs for review!

> Does the change also work with plain hostnetwork without nodeport?

Please have a look at the roster of tests:
| access mode | HostNetwork | HostPort | Result |                      Notes                       |
| :---------: | :---------: | :------: | :----: | :----------------------------------------------: |
| `Instance`  |   `true`    |  `true`  |  `ok`  |                    status quo                    |
|    `IP`     |   `true`    |  `true`  |  `ok`  |       PodIP == HostIP --> limited benefits       |
|    `IP`     |   `false`   |  `true`  |  `ok`  | PodIP != HostIP --> limited scaling, node bounde |
|    `IP`     |   `true`    | `false`  | `N/A`  |            impossible, HN implies HP             |
|    `IP`     |   `false`   | `false`  |  `OK`  |            goal: free scaling and HA             |

From my testing it is impossible to define HostNetwork but disable HostPort. [AFAIK this combination is not avail](https://kubernetes.io/docs/concepts/cluster-administration/networking/#the-kubernetes-network-model)  as the HostNetworking implies pod port mapping.

> We don't use nodeport service

Likewise, we don't either

> drawback is that we need informers and because of that client-go WATCH (internal to informers)

Curious, unaware of this issue. Wondering, as the informer is used across SIGS projects such as ClusterAutoscaler or even [suggested pattern for controller](https://github.com/kubernetes/sample-controller) - is that issue you face a corner case?

szuecs:
Thanks for the testing table, looks great to me.
Shared informers are afaik only a different api working on cache and internal using watch and hopefully Timeouts, which we can check or if not try to pass a Transport with Timeouts when we create the client.
You will observe the issue not often and only if you check "node issues" that likely everyone will just terminate because it is one way to fix it fast. For ingress (skipper, this one and external-dns), we were conservative and use things that work reliably. I think external-dns changed it or we added passing a transport to it in the early days.
If we change to informers we should change all of it to informers in my opinion.
Maybe as a follow up PR.

universam1:
>  we can check or if not try to pass a Transport with Timeouts when we create the client.

Thank you for the context, pretty impressive deep dive! Issues like this are incredible challenging to surface!
AFAIK the client defaults to https://pkg.go.dev/net/http#DefaultTransport that looks like it incorporates a timeout and H2 at least for Go 1.16!?
The rest.Client supports Timeout as well https://github.com/kubernetes/client-go/blob/4ae0d1917adff2d93ea4c6a7ae3a9e5c19280b9f/rest/transport.go#L32

Since I'm not familiar with this issue, can you help me out what the desired config would look like?


> If we change to informers we should change all of it to informers in my opinion. Maybe as a follow up PR.

IMHO a full refactor could simplify things, but might render this PR large to impossible to land.

szuecs:
> > we can check or if not try to pass a Transport with Timeouts when we create the client.
> 
> Thank you for the context, pretty impressive deep dive! Issues like this are incredible challenging to surface! AFAIK the client defaults to https://pkg.go.dev/net/http#DefaultTransport that looks like it incorporates a timeout and H2 at least for Go 1.16!? 

Interesting not sure if this helps in this case. On the other hand client-go uses H2 ping frames to detect broken connections, so should be fine.

> The rest.Client supports Timeout as well https://github.com/kubernetes/client-go/blob/4ae0d1917adff2d93ea4c6a7ae3a9e5c19280b9f/rest/transport.go#L32
> 
> Since I'm not familiar with this issue, can you help me out what the desired config would look like?

I would just start to pass a custom Transport, you can use the DefaultTransport to pass it.
When I would do it I would likely use https://pkg.go.dev/github.com/zalando/skipper/net#NewTransport.
Let's see if we observe errors and then we can work on a fix for these. Maybe H2 ping frames detect all things we need already.

> 
> > If we change to informers we should change all of it to informers in my opinion. Maybe as a follow up PR.
> 
> IMHO a full refactor could simplify things, but might render this PR large to impossible to land.

Agree out of scope, but we should have an issue that we can work on later to migrate.

universam1:
> I would likely use https://pkg.go.dev/github.com/zalando/skipper/net#NewTransport.

@szuecs struggling to leverage the skipper lib as a custom Transport here, running into the error https://github.com/kubernetes/client-go/issues/452 "using a custom transport with TLS certificate options or the insecure flag is not allowed". I don't see `TLSClientConfig` being exported to be provided for TLS setup and alternatively disabling TLS doesn't sound like an option to me. Thanks for any hint!

universam1:
@szuecs Added complete set of unit tests, implemented the Skipper NewTransport, tested in Lab, I believe this should be ready to go.
Anything missing?

go.mod
```diff
@@ -16,8 +15,11 @@ require (
...
 	github.com/sirupsen/logrus v1.8.1
 	github.com/stretchr/testify v1.7.0
+	github.com/zalando/skipper v0.13.167
 	gopkg.in/alecthomas/kingpin.v2 v2.2.6
 	gopkg.in/go-playground/assert.v1 v1.2.1 // indirect
 	gopkg.in/go-playground/validator.v9 v9.31.0 // indirect
-	gopkg.in/yaml.v2 v2.4.0 // indirect
+	k8s.io/api v0.22.0
+	k8s.io/apimachinery v0.22.0
+	k8s.io/client-go v0.22.0
```
szuecs, line 24: @mikkeloscar what do you think about kubernetes dependency ^

szuecs:
@universam1 from my side only DCO :)

szuecs:
```
Run make lint
golangci-lint run ./...
level=error msg="Running error: context loading failed: failed to load packages: timed out to load packages: context deadline exceeded"
level=error msg="Timeout exceeded: try increasing it by passing --timeout option"
```

restarted, in case it happens again please check why it takes so long and if it really takes so long increase the timeout, please

universam1:
> please check why it takes so long and if it really takes so long increase the timeout, please
It requires a few adjustments on the CI as much as I saw, fine from my side if that's desired

szuecs:
yes, please

szuecs:
:+1:

universam1:
added a section in the readme to describe that mode, anything missing @szuecs ?

README.md
```diff
@@ -590,6 +591,28 @@ By default, the controller will expose both HTTP and HTTPS ports on the load bal
...
+## AWS CNI Mode (experimental)
+
+The default operation mode of the controller (`target-access-mode=HostPort`) is to link the target groups to the autoscaling group. The target group type is `instance`, requiring the ingress pod to be accessible through a `HostNetwork` and `HostPort`.
+
+In *AWS CNI Mode* (`target-access-mode=AWSCNI`) the controller actively manages the target group members. Since AWS EKS cluster running AWS VPC CNI have their pods as first class members in the VPCs, they can receive the traffic directly, being managed through a target group type is `ip`, which means there is no necessity for the HostPort indirection.
+
+### Notes
+
+- For security reasons the HostPort requirement might be of concern
+- Direct management of the target group members is significant faster to  compared to the AWS linked mode, but it requires a running controller -od for updates. As of now, the controller is not prepared for high availability replicated setup.
```
AlexanderYastrebov, line 603: `-od` - typo?

aws/adapter.go
```diff
@@ -1009,3 +1048,53 @@ func nonTargetedASGs(ownedASGs, targetedASGs map[string]*autoScalingGroupDetails
...
+			return fmt.Errorf("unable to describe target health %w", err)
+		}
+		registeredInstances := make([]string, len(tgh.TargetHealthDescriptions))
+		for i, target := range tgh.TargetHealthDescriptions {
+			registeredInstances[i] = *target.Target.Id
+		}
+		toregister := difference(endpoints, registeredInstances)
+		if len(toregister) > 0 {
+			log.Info("Registering CNI targets: ", toregister)
+			err := registerTargetsOnTargetGroups(a.elbv2, a.TargetCNI.TargetGroupARNs, toregister)
```
AlexanderYastrebov, line 1070: This feels wrong: it registers all of `toregister` on all `a.TargetCNI.TargetGroupARNs`.
Maybe it should calculate all to register and all to de-register and call `registerTargetsOnTargetGroups/deregisterTargetsOnTargetGroups` outside of the loop. 
Or call `registerTargetsOnTargetGroups/deregisterTargetsOnTargetGroups` only for the current iteration target group.

aws/adapter.go
```diff
@@ -568,14 +593,26 @@ func (a *Adapter) UpdateTargetGroupsAndAutoScalingGroups(stacks []*Stack, proble
 		return
 	}
 
+	// split the full list into relevant TG types
+	targetTypesARNs, err := categorizeTargetTypeInstance(a.elbv2, targetGroupARNs)
```
AlexanderYastrebov, line 597: `categorizeTargetTypeInstance` fetches all target groups which we do once again inside of `updateTargetGroupsForAutoScalingGroup` which fells unaligned. 

The `targetTypesARNs` contains only exiting target groups out of `targetGroupARNs` which is not correct - the `updateTargetGroupsForAutoScalingGroup` also handles the case when TG from `targetGroupARNs` does not exist.

The proposed approach with CNI updates is different from the existing design: controller runs a loop which calls `UpdateTargetGroupsAndAutoScalingGroups` on every iteration.

I'm wondering if we could keep it simple like it is, e.g. controller calls `UpdateTargetGroupsAndAutoScalingGroups`, it fetches pods and instances and updates target groups accordingly, all synchronous just like everything else.

kubernetes/pods.go
```diff
@@ -0,0 +1,100 @@
...
+			case pod.DeletionTimestamp != nil:
+				name, exists := podEndpoints.LoadAndDelete(pod.Name)
+				if !exists {
+					return
+				}
+				log.Infof("Deleted pod: %s IP: %s", pod.Name, name)
+				queueEndpoints(&podEndpoints, endpointChan)
+
+			case pod.Status.ContainerStatuses != nil && len(pod.Status.ContainerStatuses) > 0 &&
+				pod.Status.ContainerStatuses[0].State.Running != nil && pod.Status.PodIP != "":
```
AlexanderYastrebov, line 51: Would it be better to poll a service endpoints API directly (assuming there is a service that selects proper pods) instead of pods and let k8s to update endpoints when pods get ready?

universam1: That was my first iteration, but as @szuecs reminded is that a service is normally not deployed with Skipper. Hence I refactored to use pod status. 
BTW although this looks weird, is is the kubectl way to describe the pod status especially to catch the "terminating" state

worker.go
```diff
@@ -641,3 +642,41 @@ func getCloudWatchAlarmsFromConfigMap(configMap *kubernetes.ConfigMap) aws.Cloud
...
+	log.Infoln("Starting CNI event handler")
+
+	rateLimiter := time.NewTicker(cniEventRateLimit)
+	defer rateLimiter.Stop()
+
+	epCh := make(chan []string, 10)
+	go func() {
+		log.Error(informer(ctx, epCh))
+		time.Sleep(cniEventRateLimit)
+	}()
```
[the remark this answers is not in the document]
universam1: missed to add the for loop

kubernetes/pods.go
AlexanderYastrebov: > but as @szuecs reminded is that a service is normally not deployed with Skipper

Having a service could be set as a requirement for CNI mode though

kubernetes/pods.go
```diff
@@ -0,0 +1,100 @@
...
+			select {
+			case <-resyncer.C:
+				queueEndpoints(&podEndpoints, endpointChan)
+			case <-ctx.Done():
+				return
+			}
+		}
+	}()
+	informer.Run(ctx.Done())
+	return nil
```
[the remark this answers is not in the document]
universam1: Core idea is the event based registration to Kubernetes using pod Informer that receives immediate notifications about pod changes, which allow almost zero delayed updates on the target groups.
The registration happens as soon as the pod received an IP from AWS VPC. Deregistration happens bound to the kubernetes event. 
This all lives on events. If we'd go to polling we'd loose all the idea of this PR to have instantaneous responds and updates, removing the delays that make up the problem we face with the current setup.

kubernetes/pods.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+const resyncInterval = 1 * time.Minute
+
+// PodInformer is a event handler for Pod events registered to, that builds a local list of valid and relevant pods
+// and sends an event to the endpoint channel, triggering a resync of the targets.
+func (a *Adapter) PodInformer(ctx context.Context, endpointChan chan<- []string) error {
+	if a.cniPodNamespace+a.cniPodLabelSelector == "" {
+		return fmt.Errorf("error scheduling PodInformer: namespace, service, and service port are required")
+	}
```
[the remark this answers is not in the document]
universam1: I don't see how that hurts but as you wish.

aws/adapter.go
```diff
@@ -1009,3 +1048,53 @@ func nonTargetedASGs(ownedASGs, targetedASGs map[string]*autoScalingGroupDetails
...
+// by polling the current list of targets thus creating a diff of what needs to be added and removed.
+func (a *Adapter) SetTargetsOnCNITargetGroups(endpoints []string) error {
+	if a.TargetCNI.TargetGroupARNs == nil || len(a.TargetCNI.TargetGroupARNs) == 0 {
+		return ErrNoTargetGroupARNs
+	}
+	for _, targetGroupARN := range a.TargetCNI.TargetGroupARNs {
+		tgh, err := a.elbv2.DescribeTargetHealth(&elbv2.DescribeTargetHealthInput{TargetGroupArn: &targetGroupARN})
+		if err != nil {
+			return fmt.Errorf("unable to describe target health %w", err)
+		}
```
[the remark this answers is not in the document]
universam1: To me it looks inconsistent to continue if an error happens, is the stack in update face or similar transient situation?

aws/adapter.go
```diff
@@ -1009,3 +1048,53 @@ func nonTargetedASGs(ownedASGs, targetedASGs map[string]*autoScalingGroupDetails
 
 	return nonTargetedASGs
 }
+
+// SetTargetsOnCNITargetGroups implements desired state for CNI target groups
+// by polling the current list of targets thus creating a diff of what needs to be added and removed.
+func (a *Adapter) SetTargetsOnCNITargetGroups(endpoints []string) error {
+	if a.TargetCNI.TargetGroupARNs == nil || len(a.TargetCNI.TargetGroupARNs) == 0 {
+		return ErrNoTargetGroupARNs
+	}
```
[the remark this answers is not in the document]
universam1: Feels potentially error prone but if you like to take that into account ok for me

kubernetes/pods.go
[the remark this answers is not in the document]
AlexanderYastrebov: It is about clarity and time spent eyeballing the "trick" that could be a canonical expression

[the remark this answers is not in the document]
AlexanderYastrebov: The place to check is wrong anyway, it is part of the config and we should check it once outside of the update loop

worker.go
```diff
@@ -641,3 +642,41 @@ func getCloudWatchAlarmsFromConfigMap(configMap *kubernetes.ConfigMap) aws.Cloud
...
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case targetCNIcfg.TargetGroupARNs = <-targetCNIcfg.TargetGroupCh:
+			log.Debugf("new message target groups: %v", targetCNIcfg.TargetGroupARNs)
```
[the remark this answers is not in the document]
universam1: This is the central event handler that is running once only and receives from channels the events, which is the concurrency solver.
I'm not a fan of locking as it can have hard to track performance issues, but if you want that locking over channels I can refactor.

kubernetes/pods.go
universam1: There is a principle downside of using service endpoints that they update _after_ the pod readiness probe is true, which stacks up the registration delay on top of the readiness delay. Say 50s initial readiness delay plus 30s registration delay and we are back to the original problem.
The current way registers the target at the same time as the pod is scheduled. The readiness probe is equal to the target health check, hence the pod is online receiving traffic the same time it is marked ready in the cluster.
This way we are saving a lot of time that is crucial when nodes terminate for spot interruptions

aws/adapter.go
```diff
@@ -432,6 +448,15 @@ func (a *Adapter) WithInternalDomains(domains []string) *Adapter {
 	return a
 }
 
+// WithTargetAccessMode returns the receiver adapter after defining the target access mode
+func (a *Adapter) WithTargetAccessMode(t string) *Adapter {
+	a.TargetCNI.Enabled = false
+	if t == TargetAccessModeAWSCNI {
+		a.TargetCNI.Enabled = true
+	}
```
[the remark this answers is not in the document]
universam1: linter suggests dropping the else for readability, but fine for me to change

kubernetes/pods.go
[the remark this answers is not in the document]
universam1: np, removing

aws/adapter.go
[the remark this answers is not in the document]
universam1: ok we can also range over nil, wasn't sure if that's safe

aws/adapter.go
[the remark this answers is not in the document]
universam1: ok we can support that

worker.go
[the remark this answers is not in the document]
szuecs: Everything that is triggered by informer should be safe from concurrency pov.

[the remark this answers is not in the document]
AlexanderYastrebov: It is not really obvious.
targetSetter is indeed `adapter.SetTargetsOnCNITargetGroups` that runs in this loop.
Maybe then we just need to use local variable instead of `targetCNIcfg.TargetGroupARNs` and pass it to the `targetSetter/SetTargetsOnCNITargetGroups` along with endpoints.

controller.go
```diff
@@ -444,6 +456,10 @@ func main() {
 	if err != nil {
 		log.Fatal(err)
 	}
+	if err = kubeAdapter.NewInclusterConfigClientset(ctx); err != nil {
+		log.Fatal(err)
+	}
+	kubeAdapter.WithTargetCNIPodSelector(targetCNIPodSelector)
```
szuecs, line 462: We should only do this in case `awsAdapter.TargetCNI.Enabled` or do I miss something?

kubernetes/client.go
```diff
@@ -29,7 +33,15 @@ type simpleClient struct {
 	httpClient *http.Client
 }
 
-const defaultControllerUserAgent = "kube-ingress-aws-controller"
+const (
+	defaultControllerUserAgent = "kube-ingress-aws-controller"
+
+	// aligned with https://pkg.go.dev/net/http#DefaultTransport
+	maxIdleConns        = 100
```
szuecs, line 40: maxIdleConns should be less, maybe 10. As we only deal with kubernetes api server here and we do not expect too many concurrent requests

worker.go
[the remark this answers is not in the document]
szuecs: Informers don't trigger 2 update events at the ~same time, as far as I remember at least.
I think the implementation of informer is something like WATCH, but cache it and have a single goroutine that triggers the callbacks (hello event). If this is true then my statement holds true. So if we do not add a concurrency bug ourselves (spawning a goroutine for example), then it should be safe and that's what I read in the code.
To @AlexanderYastrebov suggestions, makes sense to me.
@universam1 can you check this please?

kubernetes/client.go
universam1: makes sense, thanks

controller.go
universam1: Logically yes, it is used only when enabled. Since we have the default flag `aws.DefaultTargetCNILabelSelector` always available it can parse any time and we don't need to care for nil config.
But if you prefer I can move that behind a case?

worker.go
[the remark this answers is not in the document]
universam1: @szuecs Exactly how you described is what I understand and the mere reason I'm in favor of using client-go/SharedInformer/Cache as it solves all those concurrency and eventual consistency for us. My tests hold that true.
Some statements from https://pkg.go.dev/k8s.io/client-go@v0.23.1/tools/cache#SharedInformer
> Every query against the local cache is answered entirely from one snapshot of the cache's state. Thus, the result of a `List` call will not contain two entries with the same namespace and name
> For a given SharedInformer and client, the notifications are delivered sequentially. For a given SharedInformer, client, and object ID, the notifications are delivered in order
> A client must process each notification promptly; a SharedInformer is not engineered to deal well with a large backlog of notifications to deliver. Lengthy processing should be passed off to something else, for example through a `client-go/util/workqueue`.

In our case the chan substitutes.

worker.go
```diff
@@ -641,3 +642,42 @@ func getCloudWatchAlarmsFromConfigMap(configMap *kubernetes.ConfigMap) aws.Cloud
...
+			return
+		case cniTargetGroupARNs = <-targetCNIcfg.TargetGroupCh:
+			log.Debugf("new message target groups: %v", cniTargetGroupARNs)
+		case eps := <-epCh:
+			log.Debugf("new message endpoints: %v", eps)
+			if len(epCh) > 0 {
+				log.Debugf("skipping, %d newer messages queued", len(epCh))
+				continue
+			}
+			if cniTargetGroupARNs == nil || len(cniTargetGroupARNs) == 0 {
```
AlexanderYastrebov, line 671: No need for nil check as `len(nil) == 0`

universam1: Solely for the debug logging, then I'll move it out

aws/adapter.go
universam1: `updateTargetGroupsForAutoScalingGroup` is built with the assumption there is only one type of TG namely `instance`. However, as we got two types including `ip` we need to filter / categorize the set down respectively, done via `categorizeTargetTypeInstance`.
From that point I don't grasp how this is wrong @AlexanderYastrebov - it should leave the default flow untouched. I also hesitate to rewrite the existing functions, which might be an idea for the future though.

AlexanderYastrebov: I mean that there were several improvements like #127 or #436 over the time that made controller more resilient to the external, possibly accidental, changes. These improvements were not invented out of nothing but rather from real problems and incidents. Pre-filtering stack target groups using existing target groups wipes those improvements. 

We may argue that controller resources should not be changed externally; if we could guarantee this then codebase would be even simpler. But for now I am not convinced we should change things by accidentally filtering via `categorizeTargetTypeInstance` (i.e. we do not leave default flow untouched).

AlexanderYastrebov:
@universam1 What is your migration strategy from `HostPort` to `AWSCNI` if any? Would it be possible to migrate running cluster without downtime?

controller.go
```diff
@@ -280,13 +282,21 @@ func loadSettings() error {
 		Default("text/plain").StringVar(&denyInternalRespContentType)
 	kingpin.Flag("deny-internal-domains-response-status-code", "Defines the response status code for a request identified as to an internal domain when -deny-internal-domains is set.").
 		Default("401").IntVar(&denyInternalRespStatusCode)
+	kingpin.Flag("target-access-mode", "Target group accessing Ingress via HostPort or AWS VPC CNI. Set to ASG for HostPort access or CNI for pod direct IP access.").
+		Default(aws.TargetAccessModeHostPort).EnumVar(&targetAccessMode, aws.TargetAccessModeHostPort, aws.TargetAccessModeAWSCNI)
+	kingpin.Flag("target-cni-pod-selector", "AWS VPC CNI only. Defines the query (namespace/labelselector) for ingress pods that should be linked to target group. Format 'namespace/key=value'.").
+		Default(aws.DefaultTargetCNILabelSelector).StringVar(&targetCNIPodSelector)
```
AlexanderYastrebov, line 288: Maybe we should split this into two flags for namespace and for label for clarity.
I would also remove `target-` prefix and only use `cni-` prefix similar to `alb-` and `nlb-`.

Then we may also allow repeatable label flag (maybe as a next improvement) like
```
--cni-namespace=kube-system
--cni-label=application=skipper-ingress
--cni-label=component=ingress
```
E.g. we use the same `application=skipper-ingress` label for multiple components so without multi-label support we would need to introduce a special label only for the CNI selector.

kubernetes/pods.go
```diff
@@ -0,0 +1,97 @@
...
+	log.Infof("Watching for Pods with labelselector %s in namespace %s", a.cniPodNamespace, a.cniPodLabelSelector)
+	factory := informers.NewSharedInformerFactoryWithOptions(a.clientset, resyncInterval, informers.WithNamespace(a.cniPodNamespace),
+		informers.WithTweakListOptions(func(options *apisv1.ListOptions) { options.LabelSelector = a.cniPodLabelSelector }))
+
+	informer := factory.Core().V1().Pods().Informer()
+	informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		UpdateFunc: func(_, newResource interface{}) {
+			pod, ok := newResource.(*corev1.Pod)
+			if !ok {
+				log.Printf("cannot cast object to corev1.Pod %v", newResource)
```
AlexanderYastrebov, line 34: Could it ever be a non-pod if we use PodInformer? Lets remove if not

kubernetes/pods.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	for {
+		list, err := factory.Core().V1().Pods().Lister().List(labels.Everything())
+		if err == nil && len(list) > 0 {
+			break
+		}
+		log.Errorf("Error listing Pods with labelselector %s in namespace %s: %v", a.cniPodNamespace, a.cniPodLabelSelector, err)
+		time.Sleep(resyncInterval)
+	}
```
AlexanderYastrebov, line 70: What is the purpose of this loop? Could you add a comment maybe?

kubernetes/pods.go
```diff
@@ -0,0 +1,97 @@
...
+		defer resyncer.Stop()
+		for {
+			select {
+			case <-resyncer.C:
+				queueEndpoints(&podEndpoints, endpointChan)
+			case <-ctx.Done():
+				return
+			}
+		}
+	}()
```
AlexanderYastrebov, line 84: What is the purpose of this if `PodInformer` calls `queueEndpoints` already on every change?

worker.go
```diff
@@ -641,3 +642,38 @@ func getCloudWatchAlarmsFromConfigMap(configMap *kubernetes.ConfigMap) aws.Cloud
...
+				continue
+			}
+			// simple rate limit
+			<-rateLimiter.C
+			err := targetSetter(eps, cniTargetGroupARNs)
+			if err != nil {
+				log.Error(err)
+			}
+		}
+	}
```
AlexanderYastrebov, line 678: Maybe this should run `targetSetter` on either update of both `TargetGroupCh` and `epCh` like:
```go
	var cniTargetGroupARNs, eps []string
	for {
		select {
		case <-ctx.Done():
			return
		case cniTargetGroupARNs = <-targetCNIcfg.TargetGroupCh:
			log.Debugf("new message target groups: %v", cniTargetGroupARNs)
		case eps = <-epCh:
			log.Debugf("new message endpoints: %v", eps)
		}
		
		if len(epCh) > 0 || len(targetCNIcfg.TargetGroupCh) > 0 {
			continue
		}

		<-rateLimiter.C
		err := targetSetter(eps, cniTargetGroupARNs)
		if err != nil {
			log.Error(err)
		}
	}
```

controller.go
szuecs: yes please

controller.go
AlexanderYastrebov: I would also remove `DefaultTargetCNILabelSelector` and check that of CNI is enabled that namespace and at least one label are configured.
Also `DefaultTargetCNILabelSelector` uses `skipper-ingress` but controller is mostly agnostic to the actual ingress implementation (`skipper` string is not used anywhere but in docs and tests)

worker.go
universam1: Guess we need to prevent no-op triggers from TargetGroupCh though

kubernetes/pods.go
universam1: triggering an event to enforce desired state in case of low activity  cluster.
But in meantime I found that the informer provides that resync again, making this obsolete
👍🏻

kubernetes/pods.go
universam1: Initial test if ingress pods for given specs can be found, otherwise the informer would silently fail

kubernetes/pods.go
universam1: Examples show that, I'd not risk a nil-pointer exception here if that assumptions does not hold true

kubernetes/pods.go
szuecs: Please add this as comment to the code. Sounds like a good decision to check for safety.

kubernetes/pods.go
szuecs: I would also play safe, client-go was never great code, so safety first is the way to go.

aws/adapter.go
universam1: Excuse my ignorance and thanks for the additional context. It seems to me that the action of `categorizeTargetTypeInstance' is misunderstood here. There is no filtering happening in this function, just categorizing of the TG by types (i.e. `instance`, `ip`, `alb` etc.)
Therefor, the sum of `targetGroupARNs` equals to the sum of `targetTypesARNs[n]`.

Those mentioned improvements will still apply, because it is impossible that TG of any other type than `instance` are linked to an ASG group!
Hence it would be an error to send to `UpdateTargetGroupsAndAutoScalingGroups` a set of TG where any of them is not of type `instance`, it would simply fail. That's why we need to categorize the types here, and since this controller was ignorant of other TG types than `instance` before, this flow is not touched.

controller.go
szuecs: CNI (common network interface) is a different term (tools that implement are for example flannel, aws-vpc-cni, calico, ..) and I would not understand the flag name if changed.
I think the flag split is good, but then use:
```
--target-cni-namespace=kube-system
--target-cni-pod-label=application=skipper-ingress
--target-cni-pod-label=component=ingress
```

aws/asg.go
```diff
@@ -260,6 +260,25 @@ func describeTargetGroups(elbv2svc elbv2iface.ELBV2API) (map[string]struct{}, er
...
+		func(resp *elbv2.DescribeTargetGroupsOutput, lastPage bool) bool {
+			for _, tg := range resp.TargetGroups {
+				for _, v := range allTGARNs {
+					if v != aws.StringValue(tg.TargetGroupArn) {
+						continue
+					}
+					targetTypes[aws.StringValue(tg.TargetType)] = append(targetTypes[aws.StringValue(tg.TargetType)], aws.StringValue(tg.TargetGroupArn))
+				}
+			}
+			return true
```
szuecs, line 276: Not sure but I think you want to return `lastPage` here, because of the pagination possibility. 
What if the listing is more than a single page, then it will stop here if you return true IIRC.

szuecs: As far as I see we also do it wrong in `describeTargetGroups` @AlexanderYastrebov .

aws/adapter.go
szuecs: @AlexanderYastrebov I guess you should add a proposal here, such that we can make the change happen.

AlexanderYastrebov: > There is no filtering happening in this function, just categorizing of the TG by types

The code adds to the resulting `targetTypes` only TGs returned by the `DescribeTargetGroupsPagesWithContext` and thus does not add non-existent/externally deleted target groups which still could be referenced by stacks.
 
I am wondering if we can use stack's `targetAccessModeCNI` to split their target groups by mode and get away without API call at all.

worker.go
AlexanderYastrebov: Current implementation only updates when it receives endpoints update but not target groups update.
I guess this is the reason to have a "resyncer" that constantly sends endpoint updates https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/461#discussion_r787656253

> Guess we need to prevent no-diff triggers from TargetGroupCh though

And from endpoint updates send by resyncer (IMO its better to remove the resyncer altogether by running setter on either update)

aws/asg.go
AlexanderYastrebov: >  To stop iterating, return false from the fn function.

I assume simply returning true iterates through all pages.

aws/adapter.go
AlexanderYastrebov: > I am wondering if we can use stack's targetAccessModeCNI to split their target groups by mode and get away without API call at all.

I guess we could add mode as a CF template parameter and use it to set `aws.Stack` property in `mapToManagedStack`

controller.go
universam1: Actions:
* removing the default
* splitting up the flags
* supporting list of labels

Question: technically `labelselector` is the term, should we use that instead of `label` (which sounds more like an assignment to me)

worker.go
universam1: The purpose of the resync is to imply desired state, against user caused divergence, not internal inconsistency.
Eventually it doesn't matter which timer source retriggers the sync since it's idempotent, sounds good

controller.go
universam1: done

kubernetes/pods.go
universam1: done

kubernetes/pods.go
universam1: done

worker.go
universam1: done

aws/asg.go
universam1: not documented but from the sources its clearer: https://github.com/aws/aws-sdk-go/blob/main/service/elbv2/api.go#L2183

kubernetes/pods.go
universam1: settled

worker.go
```diff
@@ -641,3 +642,43 @@ func getCloudWatchAlarmsFromConfigMap(configMap *kubernetes.ConfigMap) aws.Cloud
...
+		if len(endpointCh) > 0 || len(targetCNIcfg.TargetGroupCh) > 0 {
+			log.Debugf("flushing, messages queued: %d:%d", len(endpointCh), len(cniTargetGroupARNs))
+			continue
+		}
+		// prevent cleanup due to startup inconsistenty, arns and endpoints can be empty but never nil
+		if cniTargetGroupARNs == nil || endpoints == nil {
+			continue
+		}
+
+		<-rateLimiter.C
```
AlexanderYastrebov, line 678: nit: This will always delay for 5s, probably not a big deal. Alternatively it could do a sleep like
```go
lastSync := time.Now()
for {
    ...
    // A negative or zero duration causes Sleep to return immediately.
    time.Sleep(cniEventRateLimit - time.Since(lastSync))
    lastSync = time.Now()
    ...
}
```

kubernetes/pods.go
```diff
@@ -0,0 +1,85 @@
...
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/tools/cache"
+)
+
+const resyncInterval = 1 * time.Minute
+
+// PodInformer is a event handler for Pod events registered to, that builds a local list of valid and relevant pods
+// and sends an event to the endpoint channel, triggering a resync of the targets.
+func (a *Adapter) PodInformer(ctx context.Context, endpointChan chan<- []string) error {
+	podEndpoints := sync.Map{}
```
AlexanderYastrebov, line 23: It seems we need to pre-populate `podEndpoints` map before launching informer, otherwise it is empty after controller restart and would trigger an update after receiving a single pod change.

kubernetes/pods.go
```diff
@@ -0,0 +1,85 @@
...
+	for {
+		list, err := factory.Core().V1().Pods().Lister().List(labels.Everything())
+		if err == nil && len(list) > 0 {
+			break
+		}
+		log.Errorf("Error listing Pods with labelselector %s in namespace %s: %v", a.cniPodNamespace, a.cniPodLabelSelector, err)
+		time.Sleep(resyncInterval)
+	}
+
+	informer.Run(ctx.Done())
```
AlexanderYastrebov, line 73: factory.Start [launches](https://github.com/kubernetes/client-go/blob/909a4591e80da995d2b2b83875b2185c5d3f2969/informers/factory.go#L128-L138) `go informer.Run(stopCh)`

kubernetes/pods.go
```diff
@@ -0,0 +1,85 @@
...
+			}
+		},
+	})
+
+	factory.Start(ctx.Done())
+	if !cache.WaitForCacheSync(ctx.Done(), informer.HasSynced) {
+		return fmt.Errorf("Timed out waiting for caches to sync")
+	}
+
+	// initial verification if ingress pods for given specs can be found, otherwise the informer would fail silently
```
AlexanderYastrebov, line 63: Maybe this can be used to pre-populate podEndpoints

kubernetes/pods.go
```diff
@@ -0,0 +1,85 @@
...
+	for {
+		list, err := factory.Core().V1().Pods().Lister().List(labels.Everything())
```
AlexanderYastrebov, line 65: Should it rather be something like
```go
factory.Core().V1().Pods().Lister().Pods(a.cniPodNamespace).List(labels.Parse(a.cniPodLabelSelector))
```

AlexanderYastrebov: I think factory already filters namespace and selector
https://github.com/kubernetes/client-go/blob/909a4591e80da995d2b2b83875b2185c5d3f2969/informers/factory.go#L240-L242

worker.go
universam1: Actually not, it will preload when not cosumed, and then the second call falls in the 5s interval

kubernetes/pods.go
universam1: factory.Start is non-blocking, required for caches to built up. Then calling informer.Run which is blocking. Thats the canonical way

kubernetes/pods.go
universam1: guarded pre-call

kubernetes/pods.go
AlexanderYastrebov: As far as I can tell, on controller restart:
- the main cycle will send a list of the relevant target groups.
- this informer will receive a change of a single pod and would publish an updated list of endpoints of size 1 since `podEndpoints` is empty.
The `SetTargetsOnCNITargetGroups` will take the difference between registered targets and this single item enpoint list and will de-register other endpoints.

The "guard" only verifies that TGs and EPs were received at least onec, but the update itself assumes that endpoint list contains all available EPs.

universam1: That case is not existing because the informer is long time in sync before the TGs are enumerated. It is flushing those first messages. There are several seconds between

universam1: I guess @AlexanderYastrebov you should take some time to explore this concept in a lab to learn how it behaves

kubernetes/pods.go
```diff
@@ -0,0 +1,85 @@
...
+				log.Infof("New discovered pod: %s IP: %s", pod.Name, pod.Status.PodIP)
+				queueEndpoints(&podEndpoints, endpointChan)
+			}
+		},
+	})
+
+	factory.Start(ctx.Done())
+	if !cache.WaitForCacheSync(ctx.Done(), informer.HasSynced) {
+		return fmt.Errorf("Timed out waiting for caches to sync")
+	}
```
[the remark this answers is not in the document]
AlexanderYastrebov: Please do not resolve until we come to a conclusion.

> listers/informers must be accessed and set up prior to starting the shared informers

In the ticket you are referencing the problem is different, they start informer and only then call `AddHandler`. 
I propose a way to postpone sending updates until cache is updated if that is needed. 

> factory.Start is non-blocking, required for caches to built up. Then calling informer.Run which is blocking. Thats the canonical way

we can wait for `ctx.Done()` after calling  `factory.Start`

```go
var sendUpdates int64 = 0
...
informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
		UpdateFunc: func(_, newResource interface{}) {
                        ...
                        // postpone sending endpoint updates until cache is ready
		        if atomic.LoadInt64(&sendUpdates) != 0 {
		    		queueEndpoints(&podEndpoints, endpointChan)
			}
			....
		},
		...
)
...
factory.Start(ctx.Done())
factory.WaitForCacheSync(ctx.Done())
atomic.StoreInt64(&sendUpdates, 1)
<-ctx.Done()
return ctx.Err()
```

kubernetes/pods.go
AlexanderYastrebov: Ok, from here https://pkg.go.dev/k8s.io/client-go/tools/cache#ResourceEventHandler I see that 
> OnUpdate is also called when a re-list happens, and it will
    get called even if nothing changed. This is useful for periodically
    evaluating or syncing something.

I guess this is the reason why we need to do 
```do
list, err := factory.Core().V1().Pods().Lister().List(labels.Everything())
```
and this is how you pre-populate `podEndpoints`.

Fine, but I still think there should be a barrier to send updated before re-list is complete as `UpdateFunc` is called per-pod so it would call `queueEndpoints` with growing endpoint list from size one to full list of available endpoint, wouldn't it?

kubernetes/pods.go
```diff
@@ -0,0 +1,85 @@
...
+func (a *Adapter) PodInformer(ctx context.Context, endpointChan chan<- []string) error {
+	podEndpoints := sync.Map{}
+
+	log.Infof("Watching for Pods with labelselector %s in namespace %s", a.cniPodLabelSelector, a.cniPodNamespace)
+	factory := informers.NewSharedInformerFactoryWithOptions(a.clientset, resyncInterval, informers.WithNamespace(a.cniPodNamespace),
+		informers.WithTweakListOptions(func(options *apisv1.ListOptions) { options.LabelSelector = a.cniPodLabelSelector }))
+
+	informer := factory.Core().V1().Pods().Informer()
+	informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		UpdateFunc: func(_, newResource interface{}) {
```
AlexanderYastrebov, line 31: Do we need `OnAdd` and `OnDelete` as well?

aws/adapter.go
AlexanderYastrebov: > please come up with an assisting code solution as it seems we are not progressing anywhere

We are not progressing on this point because the proposal changes existing flow:
1. create a stack
2. manually remove its TG
3. controler would reconcile and de-regester removed TG

With this proposal `targetTypesARNs[elbv2.TargetTypeEnumInstance]` passed to `updateTargetGroupsForAutoScalingGroup` and others will only contain existing TGs and thus removed TG would not be de-regestered from ASG.

kubernetes/pods.go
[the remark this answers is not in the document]
universam1: your logic is flawed, it skips the first event!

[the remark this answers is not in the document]
universam1: You can see in the unit test that the informer.Run is though required

kubernetes/pods.go
universam1: no, they are not useful

kubernetes/pods.go
AlexanderYastrebov: > informer is long time in sync before the TGs are enumerated

How is it ensured in the code? What if k8s API is slow and AWS API is fast?

kubernetes/pods.go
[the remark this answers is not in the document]
AlexanderYastrebov: >  it skips the first event! 
```go
...
atomic.StoreInt64(&sendUpdates, 1)
// trigger an update
queueEndpoints(&podEndpoints, endpointChan)
```

kubernetes/pods.go
AlexanderYastrebov: Do I get it right that when new skipper pod would be created e.g. due to scaling the informer will receive it via `UpdateFunc` and hence there is not need to handle `OnAdd`?

universam1: > * OnAdd is called when an object is added.

as I said this state is not of any use for us because the pod has no IP yet.

aws/adapter.go
universam1: Let me try to picture a solution: `updateTargetGroupsForAutoScalingGroup` receives the original set of `targetGroupARNs` just minus those `targetTypesARNs[elbv2.TargetTypeEnumIp]`. That should include all your dirty TGs as well like before but filtered the type IP out.

Would you be happy with that @AlexanderYastrebov ?

kubernetes/pods.go
universam1: It is a hypothetical problem, not really relevant.

Nevertheless your code does not work, the cache is not populating correctly.
Also the list is still sent empty at the start
```
17:14:08Z" level=info msg="Watching for Pods with labelselector application=skipper-ingress in namespace kube-system"
17:14:08Z" level=info msg="SENDING NEW LIST: []"
17:14:08Z" level=info msg="New discovered pod: skipper-ingress-577f5488f7-vc6lp IP: 172.16.39.224"
17:14:08Z" level=info msg="SENDING NEW LIST: [172.16.107.240 172.16.39.224]"
17:14:08Z" level=info msg="New discovered pod: skipper-ingress-577f5488f7-vv59z IP: 172.16.179.140"
17:14:08Z" level=info msg="New discovered pod: skipper-ingress-577f5488f7-5frgp IP: 172.16.107.240"
17:14:08Z" level=info msg="SENDING NEW LIST: [172.16.107.240 172.16.179.140 172.16.39.224]"
```

kubernetes/pods.go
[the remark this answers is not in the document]
universam1: doesnt work see above, lets use only one thread to discuss

aws/adapter.go
AlexanderYastrebov: That sounds better and easier than tracking target mode as CF stack parameter.
```go
func (a *Adapter) UpdateTargetGroupsAndAutoScalingGroups(stacks []*Stack, problems *problem.List) {
	instanceTargetGroupARNs, ipTargetGroupARNs, err := getTargetGroups(stacks)
	if err != nil {
		problems.Add("failed to categorize Target Type Instance: %w", err)
		return
	}

	if a.TargetCNI.Enabled {
		a.TargetCNI.TargetGroupCh <- ipTargetGroupARNs
	}

	// don't do anything if there are no target groups
	if len(instanceTargetGroupARNs) == 0 {
		return
	}
	...
}
```

> Would you be happy with that

I would be happy to accept changes that will not create incidents/work and I try to foresee as much of the problems as possible during review.

universam1: implemented

kubernetes/pods.go
universam1: @AlexanderYastrebov 
Some dirty hack could be to add a `time.Sleep(time.Second)` before the channel flusher that grants gracetime for the list to populate.

```golang
...
		case endpoints = <-endpointCh:
			log.Debugf("new message endpoints: %v", endpoints)
		}
		
		// grant gracetime for cache warm up
		time.Sleep(time.Second)

		if len(endpointCh) > 0 || len(targetCNIcfg.TargetGroupCh) > 0 {
			log.Debugf("flushing, messages queued: %d:%d", len(endpointCh), len(cniTargetGroupARNs))
			continue
		}
...
```
I don't like that but we can improve that later if we discover something smarter to catch the cold cache. Would you be ok with that?

universam1: implemented

AlexanderYastrebov: The `time.Sleep` hack still has a race with informer.
Why do you prefer it to delaying sending updates until cache is ready https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/461#discussion_r788928176 ?

kubernetes/pods.go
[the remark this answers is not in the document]
AlexanderYastrebov: > doesnt work see above

Could you elaborate what does not work or link to the comment?

The idea is that in the `AddEventHandler` we accumulate updates but only send them when `sendUpdates` is set.
We set sendUpdates only when waiting for the cache is complete and manually trigger `queueEndpoints(&podEndpoints, endpointChan)` after setting the flag:

```go
func (a *Adapter) PodInformer(ctx context.Context, endpointChan chan<- []string) error {
	podEndpoints := sync.Map{}
	var sendUpdates int64 = 0

	log.Infof("Watching for Pods with labelselector %s in namespace %s", a.cniPodLabelSelector, a.cniPodNamespace)
	factory := informers.NewSharedInformerFactoryWithOptions(a.clientset, resyncInterval, informers.WithNamespace(a.cniPodNamespace),
		informers.WithTweakListOptions(func(options *apisv1.ListOptions) { options.LabelSelector = a.cniPodLabelSelector }))

	informer := factory.Core().V1().Pods().Informer()
	informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
		UpdateFunc: func(_, newResource interface{}) {
			pod, ok := newResource.(*corev1.Pod)
			if !ok {
				log.Printf("cannot cast object to corev1.Pod %v", newResource)
				return
			}
			switch {

			case pod.DeletionTimestamp != nil:
				name, exists := podEndpoints.LoadAndDelete(pod.Name)
				if !exists {
					return
				}
				log.Infof("Deleted pod: %s IP: %s", pod.Name, name)
				if atomic.LoadInt64(&sendUpdates) != 0 {
		    			queueEndpoints(&podEndpoints, endpointChan)
				}

			case pod.Status.ContainerStatuses != nil && len(pod.Status.ContainerStatuses) > 0 &&
				pod.Status.ContainerStatuses[0].State.Running != nil && pod.Status.PodIP != "":
				if _, isStored := podEndpoints.LoadOrStore(pod.Name, pod.Status.PodIP); isStored {
					return
				}
				log.Infof("New discovered pod: %s IP: %s", pod.Name, pod.Status.PodIP)
				if atomic.LoadInt64(&sendUpdates) != 0 {
		    			queueEndpoints(&podEndpoints, endpointChan)
				}
			}
		},
	})

	factory.Start(ctx.Done())
	if !cache.WaitForCacheSync(ctx.Done(), informer.HasSynced) {
		return fmt.Errorf("Timed out waiting for caches to sync")
	}

	// 
	// I guess this comment should rather explain that listing triggers AddEventHandler 
	// 
	// initial verification if ingress pods for given specs can be found, otherwise the informer would fail silently
	for {
		list, err := factory.Core().V1().Pods().Lister().List(labels.Everything())
		if err == nil && len(list) > 0 {
			break
		}
		log.Errorf("Error listing Pods with labelselector %s in namespace %s: %v", a.cniPodNamespace, a.cniPodLabelSelector, err)
		time.Sleep(resyncInterval)
	}

	atomic.StoreInt64(&sendUpdates, 1)
	// force an initial update
	queueEndpoints(&podEndpoints, endpointChan)

	informer.Run(ctx.Done())
	return nil
}
```

kubernetes/pods.go
universam1: https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/461#discussion_r789576528 simply doesn't work, the events are generated yet by informer.Run, not from factory.Start . That's why we cannot omit the informer.Run

kubernetes/pods.go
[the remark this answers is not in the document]
universam1: please see https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/461#discussion_r789661577

kubernetes/pods.go
szuecs: I think hypothetical or not, if the problem exist we need to tackle it.
We have sometimes more than 1k Ingress objects in a cluster with an SLA that 40k is fine. This can easily lead to kubernetes slower than AWS API response. `time.Sleep` is never a solution, I would rather accept a TODO comment that needs to be addressed some day.

What if we just list endpoints once on creation to populate the cache?
One issue that could happen is that if we do not execute an update operation on cloud load balancers to update the endpoints list, then a deployment of kube-ingress-aws-controller could miss a new created endpoint while it was down.
I think the right thing would be at start to make sure we list endpoints through informer (warm up cache) and use the list to trigger updates to all cloud load balancers. After that we can rely onUpdate events.

go.mod
mikkeloscar: For me it's fine. I think it's the best way to do this.

szuecs: thanks!

kubernetes/pods.go
universam1: That's actually a smart idea to determine when is the cache equilibrium reached - I like that and it doesn't sound hard to implement...

universam1: @szuecs @AlexanderYastrebov Please check if you like this version. Using the `list` is convenient in several ways as it delivers the warm cache right from the start. The successive informer updates result in a no-op since we do anyway a map reduce!
That means, there is nothing like a warm-cache sync validation required!

The beauty is that even in a hard race we will process correctly a pod deletion right after the initial list just fine.

kubernetes/pods.go
```diff
@@ -0,0 +1,92 @@
...
+// and sends an event to the endpoint channel, triggering a resync of the targets.
+func (a *Adapter) PodInformer(ctx context.Context, endpointChan chan<- []string) (err error) {
+	podEndpoints := sync.Map{}
+
+	log.Infof("Watching for Pods with labelselector %s in namespace %s", a.cniPodLabelSelector, a.cniPodNamespace)
+	factory := informers.NewSharedInformerFactoryWithOptions(a.clientset, resyncInterval, informers.WithNamespace(a.cniPodNamespace),
+		informers.WithTweakListOptions(func(options *apisv1.ListOptions) { options.LabelSelector = a.cniPodLabelSelector }))
+
+	informer := factory.Core().V1().Pods().Informer()
+	factory.Start(ctx.Done())
```
AlexanderYastrebov, line 30: `factory.Start()` apparently [starts](https://github.com/kubernetes/client-go/blob/3255cdcd9cf0ce5ea718c97f4aea519130649aca/informers/factory.go#L127-L138) informers but we configure informer EventHandler later which seems wrong.

From this perspective `informer.Run(ctx.Done())` also feels wrong as factory.Start already runs it.

szuecs: I guess we can do what others do:
workqueue: https://github.com/kubernetes/client-go/blob/master/examples/workqueue/main.go#L128-L132
fakeclient: https://github.com/kubernetes/client-go/blob/master/examples/fake-client/main_test.go#L69-L72

The cache.WaitForCacheSync() does what we need:
https://github.com/kubernetes/client-go/blob/3255cdcd9cf0ce5ea718c97f4aea519130649aca/informers/factory.go#L141
https://github.com/kubernetes/client-go/blob/3255cdcd9cf0ce5ea718c97f4aea519130649aca/tools/cache/shared_informer.go#L254

I also asked in sig-apimachinery, if they consider this safe or if we have to do something else.

szuecs: so Kubernetes sig-apimachinery says that it's not an issue because List is used before Watch, which means there is no race.

kubernetes/pods.go
szuecs: I think we should not over complicate things see also https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/461/#discussion_r791660320
@AlexanderYastrebov do you agree?

kubernetes/pods.go
AlexanderYastrebov: @szuecs  
In another discussion https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/461#discussion_r788818501 this issue https://github.com/kubernetes/kubernetes/issues/92777#issuecomment-862559059 was brought up which is exactly about the problem we have now - configuring informer after factory start.

szuecs: Then we should just move the AddEventhandler call with body to here

universam1: It certainly makes no difference if AddEventhandler is added before or after the start. Eventually I don't care, but I can prove that even via the unit test.
@AlexanderYastrebov you're misreading the issue https://github.com/kubernetes/kubernetes/issues/92777#issuecomment-862559059, it is relevant to the place where WaitForCacheSync is placed regarding informers.Start

universam1: See the sources https://github.com/kubernetes/client-go/blob/master/tools/cache/shared_informer.go#L552-L556 that explain how a safe join happens to a running informer

szuecs: So this makes sure we get all entries from a fresh LIST, right?

universam1: Yes- so theoretically we could omit the explicit List before since we get this one implicitly by late binding, but it serves two reasons so I'd keep it

AlexanderYastrebov: > So this makes sure we get all entries from a fresh LIST, right?
> Yes- so theoretically we could omit the explicit List

Idk, I see it uses `addNotification`  but we only handle updates.

AlexanderYastrebov:
I guess we've spent enough time pondering on the proposed solution so it is good enough to be accepted.

AlexanderYastrebov:
:+1:

szuecs:
:+1:

szuecs:
@universam1 thanks and let's have follow up fixes if you detect something that does not work as great as it could.

universam1:
Thank you @szuecs  for releasing v0.12.17 - just wondering if this is really considered a bugfix release bump? 😺

szuecs:
@universam1 we increase minor version in case there is a need to change flags or something like this. We add always migration guide to the readme for minor release increase.

PR #463, opened 2022-01-26
default NLB with internal annotation should fallback to ALB

szuecs:
default NLB with internal annotation should fallback to ALB, because of https://docs.aws.amazon.com/elasticloadbalancing/latest/network/load-balancer-troubleshooting.html#intermittent-connection-failure that we also found in a small production incident

### DIFF
--- a/README.md
+++ b/README.md
@@ -344,7 +344,17 @@ spec:
           servicePort: main-port
 ```
 
-You can only select from `internet-facing` (default) and `internal` options.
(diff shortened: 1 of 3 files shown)


mikkeloscar:
This should probably also be noted in the docs/README

mikkeloscar:
:+1:

AlexanderYastrebov:
:+1:

PR #471, opened 2022-02-04
CNI mode: validate initial pod list

universam1:
There is a corner case where the initial listing of the pods happens during an ingress pod start phase. A pod without an assigned IP address might be included then, causing an ALB TG update error.

We are validating the pod statuses in the informer event handler but not yet at the initial listing. Adding the same validations there as well.

### DIFF
--- a/kubernetes/pods.go
+++ b/kubernetes/pods.go
@@ -43,7 +43,9 @@ func (a *Adapter) PodInformer(ctx context.Context, endpointChan chan<- []string)
 		time.Sleep(resyncInterval)
 	}
 	for _, pod := range podList {
-		podEndpoints.Store(pod.Name, pod.Status.PodIP)
+		if !isPodTerminating(pod) && isPodRunning(pod) {
+			podEndpoints.Store(pod.Name, pod.Status.PodIP)
+		}
 	}
 	queueEndpoints(&podEndpoints, endpointChan)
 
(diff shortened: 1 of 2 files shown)


universam1:
@AlexanderYastrebov @szuecs

kubernetes/pods.go
```diff
@@ -43,7 +43,9 @@ func (a *Adapter) PodInformer(ctx context.Context, endpointChan chan<- []string)
 		time.Sleep(resyncInterval)
 	}
 	for _, pod := range podList {
-		podEndpoints.Store(pod.Name, pod.Status.PodIP)
+		if isPodRunning(pod) {
```
AlexanderYastrebov, line 46: Can we get terminating pod here as well?

AlexanderYastrebov: Ah, nevermind, we only store running pods here

universam1:
> Add a test for it maybe?

NP, covered by TestAdapter_PodInformer and explicitly now

kubernetes/pods.go
```diff
@@ -43,7 +43,9 @@ func (a *Adapter) PodInformer(ctx context.Context, endpointChan chan<- []string)
...
+			podEndpoints.Store(pod.Name, pod.Status.PodIP)
+		}
```
AlexanderYastrebov, line 48: > Add a test for it maybe?
>> NP, covered by TestAdapter_PodInformer and explicitly now

Well, I meant the listing part - the filtering was introduced but no test was changed which means it was not covered.

universam1: added - the test should contain all conditions now

universam1:
@szuecs would be good to get this one out the door asap

AlexanderYastrebov:
:+1:

szuecs:
:+1:

PR #474, opened 2022-02-07
fabricgateway resource create cloud lbs

szuecs:
Similar to RouteGroups, FabricGateways should trigger a load balancer creation.
If one of these types do not exist as CRD in the cluster or rbac error, nothing should break.
Controller would now need a restart in case permissions are fixed.


szuecs:
Successfully tested in dev cluster.

AlexanderYastrebov:
:+1:

szuecs:
:+1:

Issue #475, opened 2022-02-08
doesn't clean up invalid target groups


universam1:
Invalid, non-existent target groups registrations on autoscaling groups are not removed, causing errors such as that scaling is blocked.

Invalid registrations can appear when CFN replaces a target group which happens due to various reasons.

Looks like a [bug that the controller keeps non-existent target groups](https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/9f2db47914ab9aa425f7232e427a9f6f5aed7de6/aws/asg.go#L175-L214) registered because it cannot determine the ownership - given that the TG is gone and thus not enumerable.
Probably this logic can be simplified to remove any invalid registration, even when it lost track of.

@AlexanderYastrebov @szuecs

szuecs:
The problem could be if you mix manual and our CF stack related Load Balancers. If you would clean up the ones that are not having our Tag, then you clean up things you do not own.

universam1:
> The problem could be if you mix manual and our CF stack related Load Balancers. If you would clean up the ones that are not having our Tag, then you clean up things you do not own.

This issue is only about cleaning up **non-existent** thus invalid target groups. I cannot think of a reason why it should not remove such broken links in any case

szuecs:
@universam1 but you link to code path which tests for ownership. If you like to add some cleanup code here you would need to check the Tag anyways, because we should only care about our resources. Other controllers need to keep track on theirs and manual created ones need to be manually maintained or you have a custom control loop that cleans up "broken things".
I don't see that this controller cleans up resource created by others and I do not believe right now that Cloudformation will create such a state. So right now it would be a won't fix, IMO.

mikkeloscar:
@universam1 @szuecs We do try to cleanup non-existing target groups: https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/9f2db47914ab9aa425f7232e427a9f6f5aed7de6/aws/asg.go#L176

If there are cases not handled by the current logic then it makes sense to handle that especially if it's is caused by the controller. We have a disconnect between target group creation/deletion (Cloudformation) and attaching to ASG which is done in the referenced controller code. As already stated checking for the ownership of target groups and checking for non-existing target groups is not the same.

universam1:
@szuecs Maybe I was too deep into the solution to explain verbose enough that this is a bug created by this controller.
[...]

PR #476, opened 2022-02-09
clean up invalid target groups

universam1:
fixes: #475

Given that there is only one TG and it is of type IP created by the stack, we are not callling `updateTargetGroupsForAutoScalingGroup` -> `detachTargetGroupsFromAutoScalingGroup`

This possible solution is to move this guard inside `updateTargetGroupsForAutoScalingGroup` - see https://github.com/zalando-incubator/kube-ingress-aws-controller/issues/475#issuecomment-1033827245

However, the  guard introduced in #141 is practically present now inside `updateTargetGroupsForAutoScalingGroup`  introduced by #436 and therefore we can simply remove it.

Tested in lab clusters.

Signed-off-by: Samuel Lang <gh@lang-sam.de>


universam1:
@mikkeloscar @szuecs

szuecs:
:+1:

PR #489, opened 2022-04-05
Update deps

mikkeloscar:
* Fix CI build for Go 1.18
* Update dependencies


mikkeloscar:
:+1:

AlexanderYastrebov:
```
./worker.go:271: (*github.com/zalando-incubator/kube-ingress-aws-controller/problem.List).Add format %w has arg r of wrong type interface{}
```
https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/c8670248ad2d178b1a48bdbb3c6c763c2ca71828/worker.go#L268-L273

mikkeloscar:
:+1:

worker.go
```diff
@@ -268,7 +268,7 @@ func doWork(
 	defer func() {
 		if r := recover(); r != nil {
 			debug.PrintStack()
-			problems.Add("panic caused by: %w", r)
+			problems.Add("panic caused by: %v", r)
```
szuecs, line 271: hmm that change is weird

mikkeloscar: This is the error:

```
./worker.go:271: (*github.com/zalando-incubator/kube-ingress-aws-controller/problem.List).Add format %w has arg r of wrong type interface{}
```

Would you propose some other formatting?

certs/provider.go
```diff
@@ -54,12 +54,6 @@ func (c *CertificateSummary) DomainNames() []string {
 	return c.domainNames
 }
 
-// ChainSize returns the number of intermediate certificates
-// in the chain
-func (c *CertificateSummary) ChainSize() int {
-	return len(c.chain.Subjects())
```
mikkeloscar, line 60: This was removed because of this `staticcheck` error:

```
Error: certs/provider.go:60:[13](https://github.com/zalando-incubator/kube-ingress-aws-controller/runs/5848683312?check_suite_focus=true#step:8:13): SA1019: c.chain.Subjects is deprecated: if s was returned by SystemCertPool, Subjects will not include the system roots. (staticcheck)
	return len(c.chain.Subjects())
	           ^
make: *** [Makefile:27: lint] Error 1

```

The method was only used in tests so it can be safely dropped.

worker.go
szuecs: hmm, I think we did a mistake in the func definition of problem.Add. I think we should have an error passed not a format string nor empty interface list.

szuecs: So let's keep your %v change and refactor it later.

szuecs:
:+1:

worker.go
AlexanderYastrebov: Maybe we may cast `recover()` result to `error`https://pkg.go.dev/builtin#recover

AlexanderYastrebov: @szuecs pointed out that `recover` result type may be anything so `%v` seems to be the right way.

AlexanderYastrebov:
:+1:

PR #491, opened 2022-04-07
Only disable 'excotic' features on NotFound/NoAccess errors

mikkeloscar:
In case we got any error listing `RouteGroups` or `FabricGateways` we would disable support assuming the resource is not available in the cluster.

This means that the feature can get disabled e.g. if the api-server is unavailable which was not the intention.

Fix the logic to only disable the features on specific errors: `NotFound` (404) and `NoAccess` (403).


szuecs:
:+1:

AlexanderYastrebov:
:+1:

PR #498, opened 2022-04-27
worker: check cloudformation stack status

AlexanderYastrebov:
Stack update may fail due to external changes, e.g. due to removal of
a referenced security group.

This change reports certain stack statuses as problems which prevents
`last_sync_timestamp_seconds` metric update that could be
used to monitor stack update failures.

### DIFF
--- a/aws/cf.go
+++ b/aws/cf.go
@@ -21,6 +21,7 @@ const (
 type Stack struct {
 	Name              string
 	status            string
+	statusReason      string
 	DNSName           string
 	Scheme            string
 	SecurityGroup     string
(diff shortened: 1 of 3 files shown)


aws/cf.go
```diff
@@ -69,6 +69,19 @@ func (s *Stack) ShouldDelete() bool {
...
+func (s *Stack) Err() error {
+	if s == nil {
+		return nil
+	}
+
+	switch s.status {
+	case cloudformation.StackStatusUpdateComplete:
+		return nil
+	}
+	return fmt.Errorf("bad status %s", s.status)
```
AlexanderYastrebov, line 82: CF stack StatusReason might be also useful if not empty

AlexanderYastrebov: Added

aws/cf.go
```diff
@@ -69,6 +70,29 @@ func (s *Stack) ShouldDelete() bool {
 	return true
 }
 
+// Err returns nil or an error describing the stack state.
+func (s *Stack) Err() error {
+	if s == nil {
+		return nil
+	}
+
+	switch s.status {
```
AlexanderYastrebov, line 79: https://github.com/aws/aws-sdk-go/blob/v1.43.34/service/cloudformation/api.go#L23779

szuecs:
:+1:

mikkeloscar:
:+1:

PR #505, opened 2022-05-05
fabricgateway: manage hostnames using external-dns custom resource

AlexanderYastrebov:
TODO:
- config flags


AlexanderYastrebov:
Not needed, new controller will translate FabricGateways into RouteGroups that are already supported.

PR #506, opened 2022-05-10
worker: run `doWork` at fixed rate

AlexanderYastrebov:
Currently controller waits for `pollingInterval` after each `doWork` run that takes `W` time long.
Work rate and metrics counter change rate is therefore equal to `1 / (pollingInterval + W)`
which makes it hard to configure alerts based on the counter change rate thresholds due to unknown `W`.

This change runs controller work cycle at fixed rate of `1 / pollingInterval`.

For the existing deployments the side effect would be reduced work cycle and increased API request rates,
e.g. for `W` of 15s and default `pollingInterval` of 30s the work cycle is reduced from 45s to 30s (-33%).

### DIFF
--- a/worker.go
+++ b/worker.go
@@ -236,6 +236,8 @@ func startPolling(
 	pollingInterval time.Duration,
 	globalWAFACL string,
 ) {
+	ticker := time.NewTicker(pollingInterval)
+	defer ticker.Stop()
 	for {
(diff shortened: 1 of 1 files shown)


AlexanderYastrebov:
Also if `W > pollingInterval` then controller would continuously run `doWork` without a delay

PR #509, opened 2022-05-19
fix #508: use ingress version 1 as default

### LINKED ISSUE #508: Is it expected that this controller doesn't work on 1.22 unless you pass in an ingress api version argument?
We just tried updating a cluster to EKS 1.22 (from 1.21) that was running 0.12.16 of this controller. As soon as we upgraded, the controller starts spamming the following message:

`time="2022-05-19T03:42:00Z" level=error msg="doWork failed to list ingress resources: failed to get ingress list: resource not found`

I try upgrading to the latest release. 0.12.36, same result. I then found this issue - https://github.com/zalando-incubator/kube-ingress-aws-controller/issues/448

Upon trying that solution:

```
image: registry.opensource.zalan.do/teapot/kube-ingress-aws-controller:latest
args:
  - "--ingress-api-version=networking.k8s.io/v1"
```

It now works properly. Is it expected or documented that, when trying to run this on 1.22, you need to explicitly pass in that argument? Seems odd....why wouldn't that just be the default?

szuecs:
fix #508


szuecs:
:+1:

mikkeloscar:
:+1:

PR #515, opened 2022-06-22
Limit ingress support to only networking.k8s.io/v1

katyanna:
Given the deprecation of `extensions/v1beta1` and `networking.k8s.io/v1beta1`,
remove support for them.


szuecs:
@katyanna please change the VERSION file to create the next minor version and explain in the readme.md section what the update to >= the new version will do.

katyanna:
I've talked with @mikkeloscar and decided for freezing this for now so other open source users can use the older versions if they need to.

szuecs:
Yeah true I also thought about that and you can change the default in kubernetes-on-aws which we likely already did.

szuecs:
@katyanna I think we can now go on and drop the v1beta1 code

katyanna:
:+1:

katyanna:
> @katyanna please change the VERSION file to create the next minor version and explain in the readme.md section what the update to >= the new version will do.

@szuecs Is this still relevant or should we merge without changing the version?

mikkeloscar:
> @szuecs Is this still relevant or should we merge without changing the version?

We should still change the version to indicate to users that this is no longer supported.

katyanna:
👍

zaklawrencea:
:+1:

PR #517, opened 2022-06-28
Build multi-arch container image

mikkeloscar:
Build multi-arch container image along with the single-arch open source image.

Additionally drop the OSX target in Makefile as `make build.local` does the right thing.


linki:
👍

PR #518, opened 2022-06-28
Used cdp-runtime/golang for prod build

mikkeloscar:
Multi-arch build (#517) for the prod-build step didn't work because it was still using the legacy CDP overlay. Change to cdp-runtime and get latest Go version as well.


mikkeloscar:
:+1:

AlexanderYastrebov:
:+1:

PR #528, opened 2022-07-25
Remove FabricGateway support

AlexanderYastrebov:
Initial support #474 was implemented as a part of an effort to make
FabricGateway a Skipper-native route format.

Since then the approach has changed and there is a plan to use a new controller
that would translate FabricGateway into equivalent RouteGroup.

Since RouteGroups are natively supported by Skipper and Ingress
Controller there is no need to keep track of the FabricGateway
loadbalancer name in the status.


szuecs:
:+1:

AlexanderYastrebov:
:+1:

Issue #438, opened 2021-10-05
Support additional TCP listeners on NLBs


jhuntwork:
I have a use case where an NLB needs 3 listeners, ports 80, 443, and 22. The picture below shows what I have set up manually and would like to achieve with kube-ingress-aws-controller.

![WithSkipper](https://user-images.githubusercontent.com/239626/136043445-ceae1047-eddb-4c2f-b73d-e377754c5300.jpg)

The ALB in the picture is already currently deployed and managed by the ingress controller. I would like to also automatically provision and manage the NLB similarly, but it needs the third pass-through TCP listener and target group.

I am willing to provide the additional features through a PR, but I have a few questions:

* What is the preferred mechanism to define additional listeners and target groups, should it just be an additional annotation on an Ingress?
* Might there ever be a use case for more than 3 listeners and should we account for that?
* As per the above diagram, what would be the correct way to instruct kube-ingress-aws-controller to configure the NLB to route the HTTP and HTTPS listeners to another controller-managed ALB which contains our cert and redirect rule, and route the SSH passthrough listener to a NodePort?

jhuntwork:
Through some initial investigations, I was trying to see if I could remove the ALB above as well, and just use an NLB. The issue is that Skipper only listens on one port, and we need a cheap/easy place to do HTTP redirects, which is what the ALB provides.

Currently it seems that there is no support to create a TG that points to another, abitrary endpoint. I'd have to figure out if we can find the listening endpoints of an ALB and set up the TG to point at that.

szuecs:
@jhuntwork the recent skipper version allows you to specify a redirect listener.

jhuntwork:
Awesome, thanks!

szuecs:
Our current implementation is based on an internal container that starts 2 skipper processes:
1) skipper redirect listen on port 9998 (https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/skipper/deployment.yaml#L56-L59 and https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/skipper/deployment.yaml#L85-L86)
2) skipper-ingress listen on port 9999 serving everything else

basically we have a docker container that starts run.sh, which has this:
```bash
if [ -n "$HTTP_REDIRECT" ]
then
	(skipper -address=:9998 \
		-support-listener='' -metrics-listener='' \
[...]

PR #529, opened 2022-07-26
feat: Advanced options for NLBs

jhuntwork:
This implements two new annotations in an attempt to cover the use
case described in #438.

The `zalando.org/aws-nlb-extra-listeners` annotation accepts a JSON
string that describes a list of extra listen/target ports to add to an
NLB. These will be routed to pods matching a specific label in the same
namespace as the ingress. As such, this depends on the AWS CNI mode
feature.

The `zalando.org/aws-nlb-cascade-http-to-alb` annotation allows the NLB to
"cascade" HTTP(S) traffic to another managed ALB. This was added to
solve two issues while testing in a live cluster:

1. When using skipper, if it is set to only accept https traffic, it
   will reject requests from an NLB that is offloading the
   SSL and passing on http traffic, which is the current default
   configuration for NLBs. Some clusters or use cases may require
   end-to-end encryption.
2. When using SSL offloading and testing with Gitlab (the use case
   described in the original issue) it became confused as to the actual
   protocol and produced incorrect redirect links which broke
   authentication. This could likely be remedied in the application
   itself, however the issue is not present when using the ALB and
   redirects are handled there. Since the ALB is already present,
   routing the traffic directly to it is a simple, reasonable choice.


szuecs:
Thanks @jhuntwork , I think because of vacations and some pressing issues we won't be able to review your PR until September.
I hope it's fine and not too much blocking!

jhuntwork:
@szuecs Thanks for the note! I think I was hoping for a quicker review than that, but I understand... vacations are important! Looking forward to the feedback when it comes.

szuecs:
@jhuntwork can you please rebase on current master, I had to drop the golint linter in order to make tests work again.

README.md
```diff
@@ -664,6 +666,37 @@ In *AWS CNI Mode* (`target-access-mode=AWSCNI`) the controller actively manages
...
+        "listenport": 22,
+        "targetport": 2222,
+        "podlabel": "application=ssh-service"
+    }
+]
+```
+
+The `podlabel` value is used to register targets in the target group associated with the listener. This depends on the
+AWS CNI Mode feature, where individual pods receive accessible IP addresses. The value is used to identify pods running
+in the same namespace as the ingress that will receive traffic from the load balancer.
```
szuecs, line 690: So the NLB would listen on `listenport` with `protocol` and forward traffic to `podlabel` selected pod IPs on port `targetport`.
If so, please add a line in the doc to make this more clear to the reader.

aws/cf.go
```diff
@@ -230,6 +254,7 @@ func createStack(svc cloudformationiface.CloudFormationAPI, spec *stackSpec) (st
 			cfParam(parameterIpAddressTypeParameter, spec.ipAddressType),
 			cfParam(parameterLoadBalancerTypeParameter, spec.loadbalancerType),
 			cfParam(parameterHTTP2Parameter, fmt.Sprintf("%t", spec.http2)),
+			cfParam(parameterCascadeParameter, fmt.Sprintf("%t", spec.cascade)),
```
szuecs, line 257: this will change the CF stack as far as I understand, so it will retrigger an update.
@mikkeloscar @AlexanderYastrebov I guess it's fine to do in this case, right?

aws/cf_template.go
```diff
@@ -120,28 +125,41 @@ func generateTemplate(spec *stackSpec) (string, error) {
...
 			Value:       cloudformation.Ref(httpsTargetGroupName).String(),
 		},
 	}
 
-	template.AddResource(httpsTargetGroupName, newTargetGroup(spec, parameterTargetGroupTargetPortParameter))
+	listener := ExtraListener{}
+	if spec.cascade {
+		listener = ExtraListener{ListenProtocol: "TCP", ListenPort: 443, TargetPort: 443, cascade: true}
+	}
+	template.AddResource(httpsTargetGroupName, newTargetGroup(spec, parameterTargetGroupTargetPortParameter, listener))
```
szuecs, line 142: can we do the code path with less changes on current CF stacks?
I would imagine it would be fine like:
```
extraListener := ExtraListener{}
if spec.cascade {
    listener = ExtraListener{ListenProtocol: "TCP", ListenPort: 443, TargetPort: 443, cascade: true}
    template.AddResource(httpsTargetGroupName, newTargetGroup(spec, parameterTargetGroupTargetPortParameter, extraListener))
} else {
    template.AddResource(httpsTargetGroupName, newTargetGroup(spec, parameterTargetGroupTargetPortParameter))
}
```
and later only use extraListener in case we have `if spec.cascade`.

aws/cf_template.go
```diff
@@ -120,28 +125,41 @@ func generateTemplate(spec *stackSpec) (string, error) {
...
 
 	if !spec.httpDisabled {
 		// Use the same target group for HTTP Listener or create another one if needed
 		httpTargetGroupName := httpsTargetGroupName
-		if spec.httpTargetPort != spec.targetPort {
+		if spec.httpTargetPort != spec.targetPort || spec.cascade {
```
szuecs, line 147: I guess it's better to do a separate if branch for readability.

szuecs:
@jhuntwork can you please split the 2 features into 2 separate PRs? It will be easier to review.
Another request is to add tests that cover the new feature in some way.

aws/cf_template.go
jhuntwork: I think you still have to pass in the parameter to the function? Which is why it's just initialized as empty first before the if statement.

jhuntwork:
> @jhuntwork can you please split the 2 features into 2 separate PRs? It will be easier to review. Another request is to add tests that cover the new feature in some way.

Sure, I can do that. I needed both to be able to handle my specific use case, but happy to split them apart. And agreed, test coverage would be good. Thanks for the feedback so far.

jhuntwork:
@szuecs This is reduced to just the first feature now and also rebased on latest master. I'm still working on a couple of tests, so those will be coming soon, hopefully.

README.md
jhuntwork: Added

jhuntwork:
@szuecs Updated with some tests. Should be ready for review again, please.

jhuntwork:
Hi @szuecs. Is there something more I need to do to get this reviewed?

szuecs:
@jhuntwork the feature enables to have git like interfaces right?
The extra listener would be for ssh for example. Do I get this right or do you see any other relevant use case?
This week is cyberweek, so likely next week. And I will try to review with @AlexanderYastrebov

sydorovdmytro:
Hi guys. Do you have any progress here? I'm also interested in this feature.

szuecs:
Hi sorry for not having progress here. We actually have some more focus on enabling this pr go forward. We want to have a better test suite that enables golden file based testing. One team member is assigned to the task and when we have it we will inform here to request to add some golden files to test the change. 
We and others had a bad incident that we don't want to have for more people in the community which would be caught by better tests.
So before adding more complexity we want to ensure safety.
I hope that's fine for everyone and I am sorry for not being proactive in our communication.

webframp:
Thanks for sharing the update @szuecs, looking forward to this getting merged eventually.

szuecs:
You can see the progress in our test strategy https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/587

szuecs:
@jhuntwork can you rebase and create a golden file test?

jhuntwork:
> @jhuntwork can you rebase and create a golden file test?

Thanks, will do!

jhuntwork:
Just coming back to this after a long while. Rebased and will now work on the goldenfile and tests.

jhuntwork:
The tests are working, but there's just a couple things more I want to refactor. I'll submit a few more changes and remove the Draft status once it's in better shape.

jhuntwork:
@szuecs I was thinking maybe some parts could use a refactor, but I might just need to get some of your input on this again now before changing anything else. Do you have time to review?

aws/adapter.go
```diff
@@ -655,8 +665,33 @@ func (a *Adapter) UpdateTargetGroupsAndAutoScalingGroups(stacks []*Stack, proble
 		a.TargetCNI.TargetGroupCh <- targetTypesARNs[elbv2.TargetTypeEnumIp]
 	}
 
+	// run through any target groups with ALB targets and register all ALBs
+	for _, tg := range targetTypesARNs[elbv2.TargetTypeEnumAlb] {
```
szuecs, line 669: What's about NLBs?

aws/adapter.go
```diff
@@ -1102,36 +1139,56 @@ func nonTargetedASGs(ownedASGs, targetedASGs map[string]*autoScalingGroupDetails
 	return nonTargetedASGs
 }
 
+func (a *Adapter) getRegisteredTargets(tgARN string) ([]string, error) {
+	tgh, err := a.elbv2.DescribeTargetHealth(&elbv2.DescribeTargetHealthInput{TargetGroupArn: &tgARN})
+	if err != nil {
+		log.Errorf("unable to describe target health %v", err)
```
szuecs, line 1145: `log.Errorf("unable to describe target health %v", err)` -> `log.Errorf("Failed to describe target health %v", err)`

aws/adapter.go
```diff
@@ -1102,36 +1139,56 @@ func nonTargetedASGs(ownedASGs, targetedASGs map[string]*autoScalingGroupDetails
 	return nonTargetedASGs
 }
 
+func (a *Adapter) getRegisteredTargets(tgARN string) ([]string, error) {
+	tgh, err := a.elbv2.DescribeTargetHealth(&elbv2.DescribeTargetHealthInput{TargetGroupArn: &tgARN})
```
szuecs, line 1143: `&tgARN` is correct but we should use `aws.String(tgARN)`

aws/adapter.go
```diff
@@ -1102,36 +1139,56 @@ func nonTargetedASGs(ownedASGs, targetedASGs map[string]*autoScalingGroupDetails
...
+	if err != nil {
+		log.Errorf("unable to describe target health %v", err)
+		return []string{}, err
+	}
+	registeredTargets := make([]string, len(tgh.TargetHealthDescriptions))
+	for i, target := range tgh.TargetHealthDescriptions {
+		registeredTargets[i] = *target.Target.Id
```
szuecs, line 1150: `*target.Target.Id` is correct but we should use `aws.StringValue(target.Target.Id)` instead

aws/adapter.go
```diff
@@ -1102,36 +1139,56 @@ func nonTargetedASGs(ownedASGs, targetedASGs map[string]*autoScalingGroupDetails
...
+	}
+	return registeredTargets, nil
+}
+
+func (a *Adapter) registerAndDeregister(new []string, old []string, tgARN string) error {
+	toRegister := difference(new, old)
+	if len(toRegister) > 0 {
+		log.Info("Registering CNI targets: ", toRegister)
```
szuecs, line 1158: better use formatstring

aws/adapter.go
```diff
@@ -1102,36 +1139,56 @@ func nonTargetedASGs(ownedASGs, targetedASGs map[string]*autoScalingGroupDetails
...
+		err := registerTargetsOnTargetGroups(a.elbv2, []string{tgARN}, toRegister)
+		if err != nil {
+			return err
+		}
+	}
+	toDeregister := difference(old, new)
+	if len(toDeregister) > 0 {
+		log.Info("Deregistering CNI targets: ", toDeregister)
```
szuecs, line 1166: better use formatstring

aws/asg.go
```diff
@@ -258,16 +258,39 @@ func describeTargetGroups(elbv2svc elbv2iface.ELBV2API) (map[string]struct{}, er
...
 				for _, v := range allTGARNs {
 					if v != aws.StringValue(tg.TargetGroupArn) {
 						continue
 					}
-					targetTypes[aws.StringValue(tg.TargetType)] = append(targetTypes[aws.StringValue(tg.TargetType)], aws.StringValue(tg.TargetGroupArn))
+					var podlabel, podnamespace string
+					log.Debugf("Looking for tags on %s", aws.StringValue(tg.TargetGroupArn))
+					out, err := elbv2svc.DescribeTags(&elbv2.DescribeTagsInput{ResourceArns: []*string{tg.TargetGroupArn}})
+					if err != nil {
+						log.Errorf("cannot describe tags on target group: %v", err)
```
szuecs, line 274: We always start uppercase for logs "Failed to describe...".

szuecs: Does it make sense to continue here, because otherwise you work with empty string below the else.

aws/cf.go
```diff
@@ -107,12 +122,15 @@ func (o stackOutput) dnsName() string {
...
+	return o[outputLoadBalancerARN]
+}
+
 func (o stackOutput) targetGroupARNs() (arns []string) {
-	if arn, ok := o[outputTargetGroupARN]; ok {
-		arns = append(arns, arn)
-	}
-	if arn, ok := o[outputHTTPTargetGroupARN]; ok {
-		arns = append(arns, arn)
+	for k, v := range o {
```
szuecs, line 130: hmm, I think it's better to have the 2 map lookups instead of a range loop over the whole map. 
Do we have more than these 2 map lookups, now?

aws/cf.go
```diff
@@ -473,6 +504,13 @@ func mapToManagedStack(stack *cloudformation.Stack) *Stack {
 		if key == ingressOwnerTag {
 			ownerIngress = value
 		}
+
+		if key == extraListenersTag {
+			decodedListeners, _ := base64.StdEncoding.DecodeString(value)
```
szuecs, line 509: you should use err and return early if there is one

aws/cf.go
```diff
@@ -473,6 +504,13 @@ func mapToManagedStack(stack *cloudformation.Stack) *Stack {
...
+			if err := json.Unmarshal(decodedListeners, &extraListeners); err != nil {
+				return &Stack{}, err
```
szuecs, line 511: should not this return nil?

aws/cf.go
```diff
@@ -497,23 +535,33 @@ func mapToManagedStack(stack *cloudformation.Stack) *Stack {
...
+					stacks = append(stacks, stack)
 				}
 			}
 			return true
 		})
 	if err != nil {
 		return nil, fmt.Errorf("findManagedStacks failed to list stacks: %w", err)
 	}
+	if len(errors) > 0 {
+		return nil, fmt.Errorf("mapToManagedStacks returned errors: %v", errors)
```
szuecs, line 563: %v to %w

aws/cf.go
```diff
@@ -497,23 +535,33 @@ func mapToManagedStack(stack *cloudformation.Stack) *Stack {
...
 	err := svc.DescribeStacksPages(&cloudformation.DescribeStacksInput{},
 		func(page *cloudformation.DescribeStacksOutput, lastPage bool) bool {
 			for _, s := range page.Stacks {
 				if isManagedStack(s.Tags, clusterID, controllerID) {
-					stacks = append(stacks, mapToManagedStack(s))
+					stack, err := mapToManagedStack(s)
+					if err != nil {
+						errors = append(errors, err)
+					}
+					stacks = append(stacks, stack)
```
szuecs, line 554: we would here either add empty stack (&Stack{}) or nil. We could think of a continue in the error path instead. 
I am not sure if we need empty stacks or nil stacks for later use, maybe for deletion or new creation.

aws/cf_test.go
```diff
@@ -196,6 +222,17 @@ func TestUpdatingStack(t *testing.T) {
 	}
 }
 
+func Test_lbARN(t *testing.T) {
+	t.Run("should return the ARN", func(t *testing.T) {
+		want := "blah"
+		s := stackOutput{outputLoadBalancerARN: want}
+		got := s.lbARN()
+		if want != got {
+			t.Errorf("unexpected result. wanted %+v, got %+v", want, got)
```
szuecs, line 231: want and got should be string so %s

kubernetes/adapter.go
```diff
@@ -210,6 +213,24 @@ func (a *Adapter) newIngress(typ IngressType, metadata kubeItemMetadata, host st
...
 	wafWebAclId, hasWAF := annotations[ingressWAFWebACLIDAnnotation]
 
+	var extraListeners []aws.ExtraListener
+	rawlisteners, hasExtraListeners := annotations[ingressNLBExtraListenersAnnotation]
+	if hasExtraListeners {
+		if loadBalancerType != loadBalancerTypeNLB {
+			return nil, errors.New("extra listeners are only supported on NLBs")
+		}
+		if err := json.Unmarshal([]byte(rawlisteners), &extraListeners); err != nil {
+			return nil, fmt.Errorf("unable to parse aws-nlb-extra-listeners annotation: %v", err)
```
szuecs, line 223: in fmt.Errorf always use %w for passing err.

kubernetes/adapter.go
```diff
@@ -210,6 +213,24 @@ func (a *Adapter) newIngress(typ IngressType, metadata kubeItemMetadata, host st
...
+		}
+		for idx, listener := range extraListeners {
+			if listener.ListenProtocol != "TCP" && listener.ListenProtocol != "UDP" && listener.ListenProtocol != "TCP_UDP" {
```
szuecs, line 226: maybe nicer to switch:
```
switch listener.ListenProtocol {
case "TCP","UDP","TCP_UDP":
// ok
default:
  return nil, error..
}
```

kubernetes/pods.go
```diff
@@ -3,28 +3,42 @@ package kubernetes
 import (
 	"context"
 	"fmt"
-	"sort"
 	"sync"
 	"time"
 
 	log "github.com/sirupsen/logrus"
+	"github.com/zalando-incubator/kube-ingress-aws-controller/aws"
```
szuecs, line 10: this should not be a dependency in the kubernetes package
We should convert it into aws.CNIEnpoint in aws/adapter.go

szuecs:
@jhuntwork I did my review. In general try to have no "aws" in "kubernetes" package and also the other way around.
We build a business object "Ingress" for the business logic for aws.

szuecs:
Maybe split the pr into more than one. One refactoring (type introduction can be one) and the actual extraListener/extraTg.

jhuntwork:
Thanks for the feedback! I'll see about making it smaller, refactoring in prep for the feature.

PR #533, opened 2022-08-22
Allow changing settings on an owned LB

mikkeloscar:
By default adding a WAF ACL ID or a security group etc. to an ingress/routegroup would require the creation of a new load balancer. This makes sense for the cases where the load balancer is shared by multiple ingresses/routegroups but for those load balancers only owned by a single resource (via `zalando.org/aws-load-balancer-shared=false`) it doesn't really make sense to create a new load balancer when the setting is just adding/changing a security group, WAF ACL or the TLS settings which can be done in-place on a running LB.

This PR handles the above case so we reduce the situations where a new load balancer needs to be created.

### DIFF
--- a/worker.go
+++ b/worker.go
@@ -88,16 +88,6 @@ func (l *loadBalancer) addIngress(certificateARNs []string, ingress *kubernetes.
 		return true
 	}
 
-	if l.ipAddressType != ingress.IPAddressType ||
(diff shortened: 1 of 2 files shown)


szuecs:
:+1:

mikkeloscar:
:+1:

PR #534, opened 2022-08-22
Update golanci-lint and address suggestions

mikkeloscar:
This updates golangci-lint to always use the latest version on github-actions. It also ~drops the `.golangci.yml` config and uses the default settings which adds new and previously suppressed linters~ updates the `.golangci.yml` to have the list of current default linters from golangci-lint.
Lastly it addresses all the lint errors identified.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -7,12 +7,10 @@ jobs:
     - uses: actions/checkout@v2
     - uses: actions/setup-go@v2
       with:
-        go-version: '^1.18'
+        go-version: '^1.19'
     - run: go version
     - run: go install github.com/mattn/goveralls@latest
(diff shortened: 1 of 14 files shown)


mikkeloscar:
:+1:

szuecs:
@mikkeloscar can we just drop golint and use our custom golangci settings?
 If you want to enable one other linter feel free to do so, but I don't want an external decision on what we run or not.

szuecs:
:+1:

mikkeloscar:
@szuecs I updates the config to include all the linters that would be enabled by default right now which also aligns with the lint errors that I fixed. Then we won't get any new linters from external, but still have a more "modern" config.

mikkeloscar:
:+1:

PR #551, opened 2022-09-20
Add skipper SecretsProvider to refresh SAToken

demonCoder95:
This uses the [`skipper/secrets`](https://github.com/zalando/skipper/tree/54ff9d009ad4e09295480f51c8d7ec0afeb2be67/secrets) module to add ability to refresh the `ServiceAccountToken` that is used by the k8s client configuration. This is required to roll out the `BoundServiceAccountTokenVolume` feature as part of [Kubernetes 1.22 upgrade](https://github.com/zalando-incubator/kubernetes-on-aws/pull/5269) where it is GA now and MUST be enabled.

Related issues:
- this repo: https://github.com/zalando-incubator/kube-ingress-aws-controller/issues/535
- pg9 issue board: https://github.bus.zalan.do/pg9/issues/issues/288

Inspiration and reference: https://github.com/zalando/skipper/pull/1295

PR #556, opened 2022-09-20
Use full path for token file

mikkeloscar:
Follow up to #551 which was missing the service account token file path when getting the token from the secretsReader.


mikkeloscar:
:+1:

szuecs:
:+1:

PR #587, opened 2023-02-28
Add goldenfile tests to ingress controller

lucastt:
The idea of this PR is to improve the confidence of developers when making changes to the ingress controler code.

The new test will fetch k8s yamls and serve it through a fake k8s API, the ingress-controler test should generate a cloud formation template to create all the necessary AWS resources for any ingress change fetched in the fake k8s API, the generated template is then compared to an expected output detailed the golden file "expected.cf".

This PR adds the boilerplate for golden tests in the ingress-controller worker level and adds two simple cases one considering the creation of an ALB and the other considering the creation of a NLB to satisfy an ingress definition in k8s.

There is more information in the body of commit messages, please take a look. :)


aws/adapter.go
```diff
@@ -494,6 +505,34 @@ func (a *Adapter) WithInternalDomainsDenyResponseContenType(contentType string)
 	return a
 }
 
+// WithCustomEc2Client returns an Adapter that will use the provided
+// Ec2 client, instead of the generic Ec2 client provided by AWS.
```
szuecs, line 509: I think it's EC2 in AWS and we should stick to the name.

aws/adapter.go
```diff
@@ -248,6 +248,17 @@ func NewAdapter(clusterID, newControllerID, vpcID string, debug, disableInstrume
 	return
 }
 
+// UpdateManifest generates a manifest again based on cluster and VPC information.
+// This method is useful when using custom AWS clients.
+func (a *Adapter) UpdateManifest(clusterID, vpcID string) (*Adapter, error) {
+	m, err := buildManifest(a, clusterID, vpcID)
+	if err == nil {
```
szuecs, line 255: I know it's saving a line of code, but please always stick to this:
- keep the good path on the left
- check always `err != nil` and return early from inside the if block

aws/adapter.go
```diff
@@ -494,6 +505,34 @@ func (a *Adapter) WithInternalDomainsDenyResponseContenType(contentType string)
...
 
+// WithCustomEc2Client returns an Adapter that will use the provided
+// Ec2 client, instead of the generic Ec2 client provided by AWS.
+func (a *Adapter) WithCustomEc2Client(c ec2iface.EC2API) *Adapter {
+	a.ec2 = c
+	return a
+}
+
+// WithCustomEc2Client returns an Adapter that will use the provided
+// Ec2 client, instead of the generic Ec2 client provided by AWS.
```
szuecs, line 516: Ec2 -> EC2

aws/fake/asg.go
```diff
@@ -0,0 +1,83 @@
...
+		m.Inputs.AttachLoadBalancerTargetGroups(m.T, input)
+	}
+	return nil, m.Outputs.AttachLoadBalancerTargetGroups.err
+}
+
+func (m *MockAutoScalingClient) DetachLoadBalancerTargetGroups(*autoscaling.DetachLoadBalancerTargetGroupsInput) (*autoscaling.DetachLoadBalancerTargetGroupsOutput, error) {
+	return nil, m.Outputs.DetachLoadBalancerTargetGroups.err
+}
+
+func MockDASGOutput(asgs ...map[string]Asgtags) *autoscaling.DescribeAutoScalingGroupsOutput {
```
szuecs, line 64: I think we can spell out DASG :)
gopls or your IDE will happily write the long name for you

aws/fake/cf.go
```diff
@@ -0,0 +1,97 @@
...
+	if out, ok := m.Outputs.DescribeStacks.response.(*cloudformation.DescribeStacksOutput); ok {
+		return out, m.Outputs.DescribeStacks.err
+	}
+	return nil, m.Outputs.DescribeStacks.err
+}
+
+func (m *MockCloudFormationClient) CreateStack(params *cloudformation.CreateStackInput) (*cloudformation.CreateStackOutput, error) {
+	m.lastGeneratedTemplate = *params.TemplateBody
+	if out, ok := m.Outputs.CreateStack.response.(*cloudformation.CreateStackOutput); ok {
+		return out, m.Outputs.CreateStack.err
```
szuecs, line 56: this is the happy path and it should be on the left for readability.
So as early wrote, please refactor all functions to have the Go style. Thanks

aws/fake/ec2.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+func (m *MockEc2Client) DescribeRouteTables(*ec2.DescribeRouteTablesInput) (*ec2.DescribeRouteTablesOutput, error) {
+	if out, ok := m.Outputs.DescribeRouteTables.response.(*ec2.DescribeRouteTablesOutput); ok {
+		return out, m.Outputs.DescribeRouteTables.err
+	}
+	return nil, m.Outputs.DescribeRouteTables.err
+}
+
+func MockDSGOutput(sgs map[string]string) *ec2.DescribeSecurityGroupsOutput {
```
szuecs, line 64: Please also spell out all `Mock<shortname><something>` to `Mock<longname><something>`

worker.go
```diff
@@ -410,6 +409,7 @@ func matchIngressesToLoadBalancers(
 	certsPerALB int,
 	ingresses []*kubernetes.Ingress,
 ) []*loadBalancer {
+	// WHy does it create a cluster local LB here?
```
szuecs, line 412: I think this can be dropped

worker.go
```diff
@@ -270,7 +270,6 @@ func doWork(
 	globalWAFACL string,
 ) (problems *problem.List) {
 	problems = new(problem.List)
-
```
szuecs, line 273: I think we can have this whitespace

worker_test.go
```diff
@@ -2,21 +2,380 @@ package main
...
+		DNSNames:     []string{"foo.bar.org"},
+		NotBefore:    time.Now(),
+		NotAfter:     time.Now().Add(time.Hour * 24),
+
+		KeyUsage:              x509.KeyUsageKeyEncipherment | x509.KeyUsageDigitalSignature | x509.KeyUsageCertSign,
+		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
+		BasicConstraintsValid: true,
+
+		IsCA: true,
+	}*/
```
szuecs, line 68: I guess we don't need commented code. In case you want to provide an example you can create an Example func, see also golang docs about that, in case you wanted to provide an example

worker_test.go
```diff
@@ -2,21 +2,380 @@ package main
...
+
+	certSummary := createDummyCertDetail(m.t, "DUMMY", []string{"foo.bar.org"}, time.Now(), time.Now().Add(time.Hour*24))
+
+	/*certificateSummary := certs.NewCertificate(
+	"foo.bar.org",
+	&caCert,
+	nil)*/
```
szuecs, line 75: cleanup commented code

worker_test.go
```diff
@@ -2,21 +2,380 @@ package main
...
+	chainKey  *rsa.PrivateKey
+	roots     *x509.CertPool
+	chainCert *x509.Certificate
+}
+
+var ca = caInfra{}
+
+func createDummyCertDetail(t *testing.T, arn string, altNames []string, notBefore, notAfter time.Time) *certs.CertificateSummary {
+	tenYears := time.Hour * 24 * 365 * 10
+	ca.Do(func() {
```
szuecs, line 91: you remember when we read this and did not see that it's a sync.Once?

worker_test.go
```diff
@@ -2,21 +2,380 @@ package main
...
+				DescribeStacks:     fake.R(nil, nil),
+				CreateStack:        fake.R(fake.MockCSOutput("42"), nil),
+			},
+			awsAdapter.LoadBalancerTypeNetwork,
+		},
+	} {
+		tt.Run(scenario.name, func(t *testing.T) {
+			b, err := os.ReadFile("./testdata/" + scenario.name + "/expected.cf")
+			if err != nil {
+				t.Error(err)
```
szuecs, line 310: Likely a Fatal (stop test) instead of Error (continue the test)

worker_test.go
```diff
@@ -2,21 +2,380 @@ package main
...
+			}
+			expected := string(b)
+
+			ec2Client := &fake.MockEc2Client{Outputs: scenario.ec2responses}
+			asgClient := &fake.MockAutoScalingClient{Outputs: scenario.asgresponses}
+			elbv2Client := &fake.MockElbv2Client{Outputs: scenario.elbv2responses}
+			cfClient := &fake.MockCloudFormationClient{Outputs: scenario.cfResponses}
+
+			a := &awsAdapter.Adapter{
+				TargetCNI: &awsAdapter.TargetCNIconfig{Enabled: false, TargetGroupCh: make(chan []string, 2)},
```
szuecs, line 320: not sure, but do we need the channel in case of `Enabled: false` ?

worker_test.go
```diff
@@ -2,21 +2,380 @@ package main
...
+			api, err := kubernetestest.NewAPI(kubernetestest.TestAPIOptions{}, f)
+			if err != nil {
+				t.Error(err)
+			}
+
+			s := httptest.NewServer(api)
+			defer s.Close()
+
+			kubeConfig := kubernetes.InsecureConfig(s.URL)
+			ingressClassFilterList := []string{}
```
szuecs, line 346: `ingressClassFilterList := make([]string, 0)`
If you know how many items you will `append`, then you can also use `ingressClassFilterList := make([]string, 0, len(items))` to allocate (`cap`) the underlying array in once.

szuecs (changes requested):
A couple of comments, nothing to big in the first read.
The test cases read really well and I am excited to see this happening.

worker_test.go
lucastt: Yeah, I'm sorry about that, I ended up missing these comments when preparing the code for review.

worker_test.go
```diff
@@ -2,21 +2,380 @@ package main
...
+
+	/*certificateSummary := certs.NewCertificate(
+	"foo.bar.org",
+	&caCert,
+	nil)*/
+	return []*certs.CertificateSummary{certSummary}, nil
+}
+
+type caInfra struct {
+	sync.Once
```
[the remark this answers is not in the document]
lucastt: Hmm, I'm not sure how to fix this. We are embedding the type `sync.Once` within the struct `caInfra`. What I can see as an option here is to give the struct a better name, indicating it is a singleton or something like that. Do you think this would address your concern?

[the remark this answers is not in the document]
lucastt: Oh, you are suggesting to not embedd this type within the struct, but instead have a `once sync.Once` field and use that instead.

worker_test.go
lucastt: Nope, you are right the chan is not needed.

worker_test.go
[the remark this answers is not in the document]
lucastt: I agree with you on the readability side but there are two problems. With a explicit once declaration whenever we create a new `caInfra` a copy of once is created making it actually not lock the ca initialization. The other problem is that the linter does not allow me. So I guess I'll just make explicit in the type name that this is a singleton, which I think is the better I can do to improve readability in this case.

aws/adapter.go
```diff
@@ -248,6 +248,17 @@ func NewAdapter(clusterID, newControllerID, vpcID string, debug, disableInstrume
 	return
 }
 
+// UpdateManifest generates a manifest again based on cluster and VPC information.
+// This method is useful when using custom AWS clients.
+func (a *Adapter) UpdateManifest(clusterID, vpcID string) (*Adapter, error) {
```
mikkeloscar, line 253: The need for this method suggest to me we are not doing this in the best way as we have both `NewAdapter` and `UpdateManifest` calling `buildManifest`. It looks like an opportunity to restructure this a bit so it's not needed. Doesn't have to be done in this PR, but it just came to my attention when looking at it.

worker_test.go
```diff
@@ -2,21 +2,344 @@ package main
...
+	vpcID := "1"
+	securityGroupID := "42"
+	running := int64(16) // See https://github.com/aws/aws-sdk-go/blob/master/service/ec2/api.go, type InstanceState
+
+	for _, scenario := range []struct {
+		name           string
+		ec2Responses   fake.Ec2MockOutputs
+		asgResponses   fake.AutoscalingMockOutputs
+		elbv2Responses fake.Elbv2MockOutputs
+		cfResponses    fake.CfMockOutputs
```
mikkeloscar, line 161: nitpick on the naming:

1. In Go we should aim to use consistent case for initialisms and acronyms. So it should be `EC2` instead of `Ec2`. Same with ELB and CF: https://github.com/golang/go/wiki/CodeReviewComments#initialisms
2. Since the package is already called `fake` the actual types may not need to be called something with `Mock`. E.g. `fake.CFOutputs` would be clear IMO.


Number 1 I would generally recommend to follow, 2 is more of a opinion so feel free to ignore that. :)

aws/adapter.go
lucastt: Yeah, I agree, I didn't refactor this at the time because I thought it was too much for this task. I created a task to do it in a next step: https://github.bus.zalan.do/pg9/issues/issues/412

lucastt:
👍

szuecs:
:+1:

lucastt:
👍

lucastt:
👍

mikkeloscar:
👍

PR #592, opened 2023-03-01
Update dependencies

mikkeloscar:
* Update build to Go 1.20
* Aggregated update of dependencies. Use version v1.22.17 for k8s.io to be in line with the version run in Zalando infrastructure.


mikkeloscar:
:+1:

PR #610, opened 2023-06-12
Replace NLBs on subnets change

mikkeloscar:
This implements an implicit feature to migrate NLBs when the subnets specified for the NLB changes.

This is done because for NLBs you can't change the subnets once the NLB has been created. This is unlike ALBs where you can freely change the subnets (and wait up to a week for it to take affect 🤷). This enables simple migration between subnets with the assumption that it's no problem to replace an NLB with a new one.

The way this works is that during the sync it checks if the existing managed NLBs have the subnets in sync. If not the NLB stack is ignored and the ingresses that were previously mapped to the NLB stack will now be mapped to a new one. The old NLB stack will become "empty" and after 1 hour (default) it will be cleaned up.

### DIFF
--- a/worker.go
+++ b/worker.go
@@ -454,7 +454,7 @@ func matchIngressesToLoadBalancers(
 			}
 
 			// Ignore NLBs with a wrong set of subnets
-			if lb.loadBalancerType == aws.LoadBalancerTypeNetwork {
+			if lb.loadBalancerType == aws.LoadBalancerTypeNetwork && lb.stack != nil {
 				subnets := subnetsByScheme(lb.scheme)
 				sort.Strings(subnets)
 				sort.Strings(lb.stack.Subnets)


worker.go
```diff
@@ -710,3 +721,25 @@ func cniEventHandler(ctx context.Context, targetCNIcfg *aws.TargetCNIconfig,
 		}
 	}
 }
+
+func equalSlices[T comparable](a, b []T) bool {
```
AlexanderYastrebov, line 725: This looks like it compares for equality ignoring order (and duplicates in a).

I am wondering if we can pull in slices helper lib and use https://pkg.go.dev/golang.org/x/exp/slices#Equal 
I think it will be part of stdlib on the next release, see https://github.com/golang/go/issues/57433 and https://pkg.go.dev/slices#Equal

If order should be ignored we can presort them.

AlexanderYastrebov: Or maybe just rename it to hint that it ignores element order

AlexanderYastrebov: I think its fine: $O(n^2)$ does not matter assuming slices are small and it does not allocate but the name could be misleading.

mikkeloscar: Changed to `slices.Equal` plus sorting first. `n` will be at most `3` so that is not a concern :)

worker_test.go
```diff
@@ -1135,6 +1137,37 @@ func TestMatchIngressesToLoadbalancers(t *testing.T) {
...
+			stack:            &aws.Stack{Subnets: []string{"a", "b", "c"}},
+		}},
+		validate: func(t *testing.T, lbs []*loadBalancer) {
+			require.Equal(t, 3, len(lbs))
+			for _, lb := range lbs {
+				if lb.clusterLocal {
+					continue
+				}
+
+				if lb.stack != nil && equalSlices[string](lb.stack.Subnets, []string{"a", "b", "c"}) {
```
szuecs, line 1163: I guess you also want to change this to `slices.Equal` as GH actions show `undefined: equalSlices`

mikkeloscar: Fixed!

mikkeloscar:
:-1:

go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/sirupsen/logrus v1.9.3
 	github.com/stretchr/testify v1.9.0
 	github.com/zalando/skipper v0.21.36
+	golang.org/x/exp v0.0.0-20231006140011-7918f672742d
```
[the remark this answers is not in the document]
mikkeloscar: updated, but the PR is still not ready, I just rebased on master.

PR #614, opened 2023-06-14
Add some more goldenfile tests

MustafaSaber:
Follow up on https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/587 for better coverage.


worker_test.go
```diff
@@ -289,19 +557,18 @@ func TestResourceConversion(tt *testing.T) {
...
 
 			a, err = a.UpdateManifest(clusterID, vpcID)
-			if err != nil {
-				t.Error(err)
-			}
-
-			f, err := os.Open("./testdata/" + scenario.name + "/ing.yaml")
-			if err != nil {
-				t.Fatal(err)
+			require.NoError(t, err)
```
MustafaSaber, line 560: Used `require` for less lines and we already use the package in other test

worker_test.go
```diff
@@ -322,18 +589,17 @@ func TestResourceConversion(tt *testing.T) {
...
+
 			log.SetLevel(log.DebugLevel)
 			problems := doWork(&mockedCertificateProvider{t: t}, 10, time.Hour, a, k, "")
-			if len(problems.Errors()) > 0 {
-				t.Error(problems.Errors())
-			}
+			require.Empty(t, problems)
+
+			// TODO: assert on number of stacks created (currently 1) to check non-shared load balancer
```
MustafaSaber, line 598: Non Shared create 2 stacks which could be known from metrics, but this require some changes to metrics, maybe we can address it in another PR.

worker_test.go
```diff
@@ -322,18 +589,17 @@ func TestResourceConversion(tt *testing.T) {
...
 
 			assert.Equal(
 				t,
-				expected,
+				strings.TrimSuffix(expected, "\n"),
```
MustafaSaber, line 602: to make git and tests happy

worker_test.go
```diff
@@ -265,14 +269,278 @@ func TestResourceConversion(tt *testing.T) {
...
+				DescribeTargetGroups: fake.R(nil, nil),
+				DescribeTags:         fake.R(nil, nil),
+			},
+			responsesCF: fake.CFOutputs{
+				DescribeStackPages: fake.R(nil, nil),
+				DescribeStacks:     fake.R(nil, nil),
+				CreateStack:        fake.R(fake.MockCSOutput("42"), nil),
+			},
+			typeLB:       awsAdapter.LoadBalancerTypeNetwork,
+			resourceType: []string{"rg", "ing"},
```
szuecs, line 538: Why do you need this?
Can't you read all yaml files in the directory?

MustafaSaber: yes that makes sense

worker_test.go
AlexanderYastrebov: It is better to trim it where `expected` is defined

lucastt:
So, when reading the expected CF templates I noticed something weird, they seem to reference values that would be received as parameters, but there is no place showing the values of these parameters. To confirm my theory I executed a vimdiff between both expected files: `testdata/ingress_rg_shared_alb/expected.cf` and `testdata/ingress_rg_notshared_alb/expected.cf` and they yield no diff, meaning they are exactly the same.

When I created the structure of these tests I didn't notice it because I just made two simple tests, on creating ALB and the other creating NLB, which yield different templates. But now I see there may be a problem in the way I built these golden file tests. 🤔 

Did I get something wrong?

testdata/ingress_alb/expected.cf
```diff
@@ -216,4 +216,4 @@
             }
         }
     }
-}
\ No newline at end of file
+}
```
szuecs, line 219: can you please cleanup these changes (do not commit these)?

MustafaSaber: Isn't the newline better for `git diff`?

MustafaSaber: @szuecs let's discuss this to see how to move forward.

lucastt:
> So, when reading the expected CF templates I noticed something weird, they seem to reference values that would be received as parameters, but there is no place showing the values of these parameters. To confirm my theory I executed a vimdiff between both expected files: `testdata/ingress_rg_shared_alb/expected.cf` and `testdata/ingress_rg_notshared_alb/expected.cf` and they yield no diff, meaning they are exactly the same.
> 
> When I created the structure of these tests I didn't notice it because I just made two simple tests, on creating ALB and the other creating NLB, which yield different templates. But now I see there may be a problem in the way I built these golden file tests. 🤔
> 
> Did I get something wrong?

Created a new issue for this. https://github.com/zalando-incubator/kube-ingress-aws-controller/issues/626

testdata/ingress_rg_notshared_alb/k8s/rg.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: zalando.org/v1
+kind: RouteGroup
+metadata:
+  name: my-route-group
+  annotations:
+    zalando.org/aws-load-balancer-shared: "false"
```
lucastt, line 6: This is still not showing up in the tests. Looks like the test is not capturing this detail. I'm not entirely sure if this is something that is not being tested or a configuration error.

testdata/ingress_rg_shared_alb/k8s/rg.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: zalando.org/v1
+kind: RouteGroup
+metadata:
+  name: my-route-group
+  annotations:
+    zalando.org/aws-load-balancer-shared: "false"
```
lucastt, line 6: I believe this should be true?

testdata/ingress_rg_notshared_alb/k8s/rg.yaml
lucastt: Seems to be a problem with the test.

testdata/ingress_rg_notshared_alb/template.cf
```diff
@@ -0,0 +1,219 @@
...
+                    {
+                        "CertificateArn": "DUMMY"
+                    }
+                ],
+                "ListenerArn": {
+                    "Ref": "HTTPSListener"
+                }
+            }
+        },
+        "LB": {
```
lucastt, line 126: This template is exactly the same as `ingress_rg_shared_alb`, this is not what I expected. Since there is an ingress and route_group configured and they should not share the same LB I would expect two LBs in the template. Did I get something wrong?

lucastt: This comment from Mustafa seems to be the answer to my question: https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/614#discussion_r1229775130

lucastt: The only weird thing is that if it creates two stacks, this stacks should have different tags, one LB should not be shared and the other LB should be sharable. From what I understood this difference should be reflected by the tags, something like `"ingress:owner":"$RGName"` in the LB/stack that is not sharable.

This means I would have to identify which stack is from which k8s object and see if the outputs match with the expected response. I still don't know how to do that.

lucastt:
👍

testdata/ingress_rg_notshared_alb/input/k8s/01-ing.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: myingress
```
szuecs, line 4: Should not this be also `shread: false` ?, at least the directory tells about "notshared".

testdata/ingress_rg_shared_alb/input/k8s/01-ing.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: myingress
```
szuecs, line 4: wha does the routegroup has `    zalando.org/aws-load-balancer-shared: "true"` but ingress not?

testdata/ingress_rg_shared_nlb/input/k8s/01-ing.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: myingress
```
szuecs, line 4: Same here:    zalando.org/aws-load-balancer-shared: "true"

szuecs:
@MustafaSaber the changes are great progress for tests!
If you do nonshared / share, please keep routegroup in ingress with annotations in-sync.
You can change the .kube file (at least in skipper tests) to configure the defaults of the controller and then test defaults, too. From my pov, to merge the only thing to make sure is that the tests are consistent in their annotation use.

testdata/ingress_rg_shared_alb/input/k8s/01-ing.yaml
lucastt: I don't know the original idea but in theory it should not make any difference since if not specified the ingress-controller will create only one LB for both RG and ingress.

I don't have strong opinions about this, I would not mind removing it or letting it there to make it more explicit.

lucastt: I'll make both have the same annotations.

testdata/ingress_rg_notshared_alb/input/k8s/01-ing.yaml
lucastt: I think here the idea is to have one configured only with default values and another with `shared: false` and test if the ingress controller would create two LBs, one for the ingress that shares the LB and other to a RG that does not share the LB.

Maybe we could improve this test naming.

MustafaSaber:
Thanks @lucastt for taking over 🙏!

aws/fake/cf.go
```diff
@@ -80,9 +86,8 @@ func MockCSOutput(stackId string) *cloudformation.CreateStackOutput {
 }
 
 func (m *CFClient) UpdateStack(params *cloudformation.UpdateStackInput) (*cloudformation.UpdateStackOutput, error) {
-	m.lastStackTags = params.Tags
-	m.lastStackParams = params.Parameters
-	m.lastStackTemplate = *params.TemplateBody
+	// TODO: Update stack needs another logic to register state update, so createStack and updateStack don't
+	// mess with each other states.
```
szuecs, line 90: is this a TODO for the next PR or should be cleaned up?

testdata/ing_shared_rg_notshared_alb/input/k8s/01-ing.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: myingress
+  annotations:
+    zalando.org/aws-load-balancer-shared: "true"
```
szuecs, line 6: test name is called "ing_shared_.." which suggests that it should be `shared: false`

aws/fake/cf.go
lucastt: I don't have a specific next PR planned but I believe this is a useful reminder for future test implementations.

This is an area that is not tested yet. We only test creation we still don't test updates. The reason I left the TODO there is to remind in next iterations that update should not use the same variables create uses to store state to avoid bugs.

testdata/ing_shared_rg_notshared_alb/input/k8s/01-ing.yaml
lucastt: The ingress is shared, the RG not shared. At least that was my intent with the naming. :S

The idea of this test is to emulate a scenario where we have a ingress that shares an LB with other ingresses and RouteGroups and a RouteGroup with a special case that does not wish to share the LB. That is why one of them is `true` and the other is `false`.

I know it can be confusing, but I also don't have better ideas for the name of this test. Do you have any suggestions?

aws/fake/cf.go
szuecs: Ok, thanks!
Maybe create an issue and link from the TODO to the issue?

testdata/ing_shared_rg_notshared_alb/input/k8s/01-ing.yaml
szuecs: So basically your test run uses one ingress and one routegroup and compares what happens if you have on "not shared" and the other "shared". Ok, now it makes sense to me.

aws/fake/cf.go
lucastt: Done: https://github.com/zalando-incubator/kube-ingress-aws-controller/issues/653

lucastt:
👍

aws/fake/cf.go
```diff
@@ -86,8 +86,9 @@ func MockCSOutput(stackId string) *cloudformation.CreateStackOutput {
 }
 
 func (m *CFClient) UpdateStack(params *cloudformation.UpdateStackInput) (*cloudformation.UpdateStackOutput, error) {
-	// TODO: Update stack needs another logic to register state update, so createStack and updateStack don't
-	// mess with each other states.
+	// TODO: https://github.com/zalando-incubator/kube-ingress-aws-controller/issues/653
+    // Update stack needs to use different variable to register change history, 
+    // so createStack and updateStack mocks don't mess with each other states.
```
szuecs, line 91: indentation looks wrong, can you check with go fmt to rewrite it?

lucastt: Yes, I'm sorry!

szuecs:
:+1:

PR #624, opened 2023-07-01
Fix security group auto-discovery filter

### LINKED ISSUE #623: Inaccurate set of filters to describe security groups
It seems that the current set of filters that used for `DescribeSecurityGroups` is inaccurate. The mentioned code passes the tests and works fine in general case, but can lead to wrong security group discovery in complex environments (when different `controller-id` values are used to separate `kube-ingress-aws-controller` instances).

To give you an example. There are two security groups with tags `kubernetes:application=controller` and `kubernetes:application=controller-backup`. With the current set of filters discovery will work like "contains" filter, but not an "exact" one. 

https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/9f9d7c1bc57066f0350a0a684f92dd7d5f3ec45b/aws/ec2.go#L283-L310

In our example case, instance with controller-id `controller` will discover both security groups and might choose the wrong one. My personal observation is that the AWS API returns the list of found security groups in unsorted order. The behaviour is reproduced constantly.

https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/9f9d7c1bc57066f0350a0a684f92dd7d5f3ec45b/aws/ec2.go#L321

Here is an example snippet from AWS SDK for Go [documentation](https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#EC2.DescribeSecurityGroups) with reference filters usage.

```golang
Filters: []*ec2.Filter{
        {
            Name: aws.String("tag:<tagName>"),
            Values: []*string{
                aws.String("<tagValue>"),
            },
        },
    },
```

antonsukhanov:
Resolves #623

### DIFF
--- a/aws/ec2.go
+++ b/aws/ec2.go
@@ -283,25 +283,13 @@ func findSecurityGroupWithClusterID(svc ec2iface.EC2API, clusterID string, contr
 	params := &ec2.DescribeSecurityGroupsInput{
 		Filters: []*ec2.Filter{
 			{
-				Name: aws.String("tag-key"),
-				Values: []*string{
-					aws.String(clusterIDTagPrefix + clusterID),
-				},
-			},
-			{
-				Name: aws.String("tag-value"),
+				Name: aws.String("tag:" + clusterIDTagPrefix + clusterID),
 				Values: []*string{
 					aws.String(resourceLifecycleOwned),
 				},
 			},
 			{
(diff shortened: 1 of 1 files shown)


AlexanderYastrebov:
Thank you.

AlexanderYastrebov:
:+1:

szuecs:
:+1:

Issue #626, opened 2023-07-13
Golden file tests are blind to some parameter changes that are not ALB/NLB selection


lucastt:
Problem statement:

Expected CF templates reference values that would be received as parameters, but the expected values are never included in the expected template, thus the expected template can be always the same as long as it respect if its a ALB or NLB template. Essentially, this would not test all other parameters except by the LB type.

For more details see https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/614#issuecomment-1598304327

---

Possible solution:

Currently the output compared to the expected template are the ones yield by [`aws.generateTenplate(...)`](https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/53a7796d913818f7a0e550af5d5dba903c71a240/aws/cf_template.go#L40C19-L40C19). We need to also consider the input parameters set [here](https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/53a7796d913818f7a0e550af5d5dba903c71a240/aws/cf.go#L218).

PR #649, opened 2023-10-18
Make Resource conversion tests also check params and tags

lucastt:
This changes should fix the current behavior of golden file tests that do not really check important parameters and only check the correctness of the template, which is not really testing the stack generation algorithm of the ingress controller.

https://github.com/zalando-incubator/kube-ingress-aws-controller/issues/626


aws/fake/cf.go
```diff
@@ -52,7 +62,10 @@ func (m *CFClient) DescribeStacks(in *cloudformation.DescribeStacksInput) (*clou
 }
 
 func (m *CFClient) CreateStack(params *cloudformation.CreateStackInput) (*cloudformation.CreateStackOutput, error) {
-	m.lastGeneratedTemplate = *params.TemplateBody
+	m.lastStackTags = params.Tags
+	m.lastStackParams = params.Parameters
+	m.lastStackTemplate = *params.TemplateBody
```
AlexanderYastrebov, line 67: Could params.TemplateBody be nil? Maybe its safer to use pointer as well.

AlexanderYastrebov: Ah, I see it was dereferenced in the existing code.

AlexanderYastrebov:
:+1:

AlexanderYastrebov:
Marked as do-not-merge until we decide if we merge to master or a feature branch

aws/fake/cf.go
lucastt: It should not be possible, if this is the case this should be an error.

szuecs:
:+1:

szuecs:
@AlexanderYastrebov I think in this case there is only test changed, so no production code path can change so we can also merge it to master IMO.

Issue #652, opened 2023-10-26
Add an option to allow to have only tagged certs to be available for use


szuecs:
As of today all `issued` ACM and IAM certificates are detected and possibly put into the TLS Listener of ALB/NLB dependeing on matching hostnames and ingress/routegroup resources.
After a production incident, one idea is to make the switch of certificates more explicit. 
To allow explicitly taking a certificate to production, we can have a flag `--tag=k=v` that would only detect certificates that have a tag key `k` with a tag value `v`. This option should be optional, which ensures non breaking change and deployments can ensure the migration, before using this feature.

szuecs:
this was done

PR #658, opened 2023-11-03
Add option to filter certificates by tag before adding it to LB

lucastt:
This feature aims to enable `kube-ingress-aws-controller` to ignore certificates based on tags. Effectively what this means is, you can specify a tag through the command line and the ingress controller will only use IAM/ACM certificates that have that specific tag when matching the host to define which certificate is more appropriate to add to the LB.

If a tag is not specified the ingress controller shall pick all available certificates for matching.

The main idea behind this feature is to make it very specific which certificates should be considered, this feature does not avoid incidents, but make it clearer to operators which certificates are being considered in the matching phase.

This PR addresses the issue: https://github.com/zalando-incubator/kube-ingress-aws-controller/issues/652

### DIFF
--- a/aws/acm.go
+++ b/aws/acm.go
@@ -2,6 +2,7 @@ package aws
 
 import (
 	"crypto/x509"
+	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/acm"
@@ -10,16 +11,17 @@ import (
 )
 
 type acmCertificateProvider struct {
-	api acmiface.ACMAPI
+	api       acmiface.ACMAPI
+	filterTag string
 }
 
-func newACMCertProvider(api acmiface.ACMAPI) certs.CertificatesProvider {
-	return &acmCertificateProvider{api: api}
+func newACMCertProvider(api acmiface.ACMAPI, certFilterTag string) certs.CertificatesProvider {
+	return &acmCertificateProvider{api: api, filterTag: certFilterTag}
 }
 
 // GetCertificates returns a list of AWS ACM certificates
 func (p *acmCertificateProvider) GetCertificates() ([]*certs.CertificateSummary, error) {
-	acmSummaries, err := getACMCertificateSummaries(p.api)
+	acmSummaries, err := getACMCertificateSummaries(p.api, p.filterTag)
 	if err != nil {
 		return nil, err
 	}
@@ -34,18 +36,45 @@ func (p *acmCertificateProvider) GetCertificates() ([]*certs.CertificateSummary,
 	return result, nil
 }
 
-func getACMCertificateSummaries(api acmiface.ACMAPI) ([]*acm.CertificateSummary, error) {
+func getACMCertificateSummaries(api acmiface.ACMAPI, filterTag string) ([]*acm.CertificateSummary, error) {
 	params := &acm.ListCertificatesInput{
 		CertificateStatuses: []*string{
 			aws.String(acm.CertificateStatusIssued),
 		},
 	}
 	acmSummaries := make([]*acm.CertificateSummary, 0)
+
 	err := api.ListCertificatesPages(params, func(page *acm.ListCertificatesOutput, lastPage bool) bool {
 		acmSummaries = append(acmSummaries, page.CertificateSummaryList...)
 		return true
 	})
+
+	if tag := strings.Split(filterTag, "="); len(tag) == 2 {
+		return filterCertificatesByTag(api, acmSummaries, tag[0], tag[1])
+	}
+
 	return acmSummaries, err
+}
+
+func filterCertificatesByTag(api acmiface.ACMAPI, allSummaries []*acm.CertificateSummary, key, value string) ([]*acm.CertificateSummary, error) {
+	prodSummaries := make([]*acm.CertificateSummary, 0)
+	for _, summary := range allSummaries {
+		in := &acm.ListTagsForCertificateInput{
+			CertificateArn: summary.CertificateArn,
+		}
+		out, err := api.ListTagsForCertificate(in)
+		if err != nil {
+			return nil, err
+		}
+
+		for _, tag := range out.Tags {
+			if *tag.Key == key && *tag.Value == value {
+				prodSummaries = append(prodSummaries, summary)
+			}
+		}
+	}
+
+	return prodSummaries, nil
 }
 
 func getCertificateSummaryFromACM(api acmiface.ACMAPI, arn *string) (*certs.CertificateSummary, error) {
--- a/aws/acm.go
+++ b/aws/acm.go
@@ -49,7 +49,7 @@ func getACMCertificateSummaries(api acmiface.ACMAPI, filterTag string) ([]*acm.C
 		return true
 	})
 
-	if tag := strings.Split(filterTag, "="); len(tag) == 2 {
+	if tag := strings.Split(filterTag, "="); filterTag != "=" && len(tag) == 2 {
 		return filterCertificatesByTag(api, acmSummaries, tag[0], tag[1])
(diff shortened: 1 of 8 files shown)


lucastt:
👍

aws/acm.go
```diff
@@ -34,20 +36,47 @@ func (p *acmCertificateProvider) GetCertificates() ([]*certs.CertificateSummary,
...
 		},
 	}
 	acmSummaries := make([]*acm.CertificateSummary, 0)
+
 	err := api.ListCertificatesPages(params, func(page *acm.ListCertificatesOutput, lastPage bool) bool {
 		acmSummaries = append(acmSummaries, page.CertificateSummaryList...)
 		return true
 	})
+
+	if tag := strings.Split(filterTag, "="); len(tag) == 2 {
```
MustafaSaber, line 52: if filterTag = "=" this condition will also work.

RomanZavodskikh: Is this case possible?

aws/acm.go
```diff
@@ -34,20 +36,47 @@ func (p *acmCertificateProvider) GetCertificates() ([]*certs.CertificateSummary,
...
+		return filterCertificatesByTag(api, acmSummaries, tag[0], tag[1])
+	}
+
 	return acmSummaries, err
 }
 
+func filterCertificatesByTag(api acmiface.ACMAPI, allSummaries []*acm.CertificateSummary, key, value string) ([]*acm.CertificateSummary, error) {
+	prodSummaries := make([]*acm.CertificateSummary, 0)
```
RomanZavodskikh, line 60: minor: do you expected (in most cases) prodSummaries to have zero length or to have length of allSummaries?
In first case, 0 is right choice. In the second case, len(allSummaries) should be instead of 0.

aws/acm_test.go
```diff
@@ -69,9 +71,64 @@ func TestACM(t *testing.T) {
...
+			),
+			filterTag: "production=true",
+			expect: acmExpect{
+				ARN:         "foobar",
+				DomainNames: []string{"foobar.de"},
+				Error:       nil,
+			},
+		},
+		{
+			msg: "Found ACM Cert with incorrect filter tag",
```
RomanZavodskikh, line 102: minor: It seems ACM Cert **was not** found in this case.

aws/acm.go
MustafaSaber: Maybe we don't need to handle I'm just mentioning the case

aws/acm.go
lucastt: Regarding the initial size of the array I don't believe this influences functionality or performance in any way. Functionality because whenever we append we will allocate space. Performance because the time spent in allocation is a lot smaller than time spent on the network call.

aws/acm.go
lucastt: True, I'll fix that.

szuecs: What is the comment about @MustafaSaber ?
I don't understand your suggestion.

aws/iam_test.go
```diff
@@ -193,3 +194,87 @@ func TestIAMParseError(t *testing.T) {
 	_, err := provider.GetCertificates()
 	require.Equal(t, ErrNoCertificates, err)
 }
+
+func TestIAMTagFiltering(t *testing.T) {
+	foobarNotBefore := time.Date(2017, 3, 29, 16, 11, 32, 0, time.UTC)
+	foobarNotAfter := time.Date(2027, 3, 27, 16, 11, 32, 0, time.UTC)
```
szuecs, line 200: 2027 -> 3027 ?

szuecs:
lgtm

aws/iam_test.go
lucastt: Yeah, its getting close hehe

lucastt: It seems this is not a problem, we only evaluate if the cert is still valid in the matching phase. These is another test in the same file with `NotAfter ~= 2018`, so I guess its fine.

lucastt:
👍

aws/acm.go
```diff
@@ -34,20 +36,47 @@ func (p *acmCertificateProvider) GetCertificates() ([]*certs.CertificateSummary,
...
 		},
 	}
 	acmSummaries := make([]*acm.CertificateSummary, 0)
+
 	err := api.ListCertificatesPages(params, func(page *acm.ListCertificatesOutput, lastPage bool) bool {
 		acmSummaries = append(acmSummaries, page.CertificateSummaryList...)
 		return true
 	})
+
+	if tag := strings.Split(filterTag, "="); filterTag != "=" && len(tag) == 2 {
```
RomanZavodskikh, line 52: minor: I think for readability it's better to create two if's with the same return here.
The first if will check filterTag, the second will split and check len.
With such two if's, the `filterTag != "="` won't be lost somewhere in the middle of the line and would be much easier to notice.

aws/acm_test.go
```diff
@@ -69,9 +71,64 @@ func TestACM(t *testing.T) {
 				Error:       nil,
 			},
 		},
+		{
+			msg: "Found ACM Cert with correct filter tag",
```
RomanZavodskikh, line 75: Would you like to add tests with several certificates in the list.
For example, "2 certs with corresponding tag" or "1 cert with corresponding tag + 1 cert with not corresponding tag"?

controller.go
```diff
@@ -115,6 +116,9 @@ func loadSettings() error {
 		StringMapVar(&additionalStackTags)
 	kingpin.Flag("cert-ttl-timeout", "sets the timeout of how long a certificate is kept on an old ALB to be decommissioned.").
 		Default(defaultCertTTL).DurationVar(&certTTL)
+
+	kingpin.Flag("cert-filter-tag", "sets a tag so the ingress controller only consider ACM or IAM certificates that have this tag set when adding a certificate to a load balancer.").
```
RomanZavodskikh, line 120: I don't quite understand, what value for this flag should be provided, if user of the controller would like to have all certificates, without any filters?

aws/acm.go
MustafaSaber: It's just an edge case, that someone will set filterTag to be `"="`

szuecs: ah got it, but then you have not 2 and it's fine to not support `=` as filter tag nor as value

aws/acm.go
szuecs: I don't see how `filterTag != "=" ` can help in the condition. I think `len(tag)==2` is enough, because if  `filterTag == "=" ` then it will have 3 parts after split.

szuecs: Ah https://go.dev/play/p/oGRkFVn_zu9 now I got it :D

aws/iam_test.go
szuecs: ah we also have a fixed cert so all data is fixed, true.

controller.go
szuecs: You don't set it -> nil -> no tag filter is used.

aws/acm.go
```diff
@@ -34,20 +36,47 @@ func (p *acmCertificateProvider) GetCertificates() ([]*certs.CertificateSummary,
...
 			aws.String(acm.CertificateStatusIssued),
 		},
 	}
 	acmSummaries := make([]*acm.CertificateSummary, 0)
+
 	err := api.ListCertificatesPages(params, func(page *acm.ListCertificatesOutput, lastPage bool) bool {
 		acmSummaries = append(acmSummaries, page.CertificateSummaryList...)
 		return true
 	})
+
```
szuecs, line 51: maybe we should add a
```
if filterTag = "" {
return acmSummaries
}
```

lucastt:
👍

szuecs:
:+1:

PR #659, opened 2023-11-20
aws: log when multiple subnets found in the same AZ

AlexanderYastrebov:
-


aws/adapter.go
```diff
@@ -948,9 +948,12 @@ func (a *Adapter) FindLBSubnets(scheme string) []string {
 
 		// If we have two subnets for the same AZ we arbitrarily choose
 		// the one that is first lexicographically.
-		if strings.Compare(existing.id, subnet.id) > 0 {
+		if subnet.id < existing.id {
 			subnetsByAZ[subnet.availabilityZone] = subnet
 		}
+
+		log.Warnf("Found subnets %s and %s for availability zone %s, using %s",
+			existing.id, subnet.id, subnet.availabilityZone, subnetsByAZ[subnet.availabilityZone])
```
AlexanderYastrebov, line 956: This is not precise as there could be more then two sunbets with the tag and the final subnet is only known after loop ends

AlexanderYastrebov: Fixed

szuecs:
:+1:

AlexanderYastrebov:
This makes sense to have only with cluster id filter
https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/6780b71333d9cbda3ca7601e4df4a25e50005013/aws/ec2.go#L193-L207

PR #667, opened 2023-12-21
Add test case for stack in RollbackInProgress state

MustafaSaber:
If a CloudFormation stack is in `ROLLBACK_IN_PROGRESS` it returns no output, hence no target groups. While CloudFormation is working to get the stack back the controller may send traffic to LBs without TG.

This PR is trying to replicate this behaviour, current state means that we also collect stacks in this state which we shouldn't do.

### DIFF
--- a/aws/cf_test.go
+++ b/aws/cf_test.go
@@ -530,6 +530,41 @@ func TestFindManagedStacks(t *testing.T) {
 			wantErr: false,
 		},
 		{
+			name: "successfull-call-with-rollback-status",
+			given: fake.CFOutputs{
+				DescribeStackPages: fake.R(nil, nil),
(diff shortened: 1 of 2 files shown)


szuecs:
You test the current state as far as I understand. If there is an error we will get into ROLLBACK state, but do you get from that test?
I also don't get your description: 
> Trying to replicate detached target groups

How does your PR achieve this? 
I don't see it

I think what we want is to test for example a known wrong state (from our current alerting for example wrong WAF association) and then how we can come out of this state without delete/recreate (causes downtime), but maybe:
1. create new stack
2. delete broken stack

The order is important and delete needs to be postponed until some time in the future as we normally do. 
Of course this requires a much bigger test, not sure if that's easily possible.

szuecs:
:+1:

PR #671, opened 2024-01-02
Add test case for rolling in progress stacks

MustafaSaber:
If a CloudFormation stack is in `ROLLBACK_IN_PROGRESS` it returns no output, hence no target groups. While CloudFormation is working to get the stack back the controller may send traffic to LBs without TG.

This PR is trying to replicate this behaviour, current state means that we also collect stacks in this state which we shouldn't do.


szuecs:
:+1:

MustafaSaber:
👍

PR #672, opened 2024-01-03
Ingore stacks in `ROLLBACK_IN_PROGRESS` state

MustafaSaber:
Signed-off-by: Mustafa Abdelrahman <mustafa.abdelrahman@zalando.de>


MustafaSaber:
Ignoring may lead to same problem since we also don't have the output which may lead also to detaching of target groups.

PR #673, opened 2024-01-04
target groups: preserve last used arns in case of `ROLLBACK_IN_PROGRESS` state

MustafaSaber:


aws/adapter.go
```diff
@@ -61,6 +61,7 @@ type Adapter struct {
 	obsoleteInstances           []string
 	stackTerminationProtection  bool
 	stackTags                   map[string]string
+	stackLastTargerGroupARNs    map[string][]string
```
szuecs, line 64: stackLastTargerGroupARNs  -> stackLastTargetGroupARNs

aws/adapter.go
```diff
@@ -626,13 +628,29 @@ func (a *Adapter) SecurityGroupID() string {
...
+	stacks, err := findManagedStacks(a.cloudformation, a.ClusterID(), a.controllerID, a.stackLastTargerGroupARNs)
 	if err != nil {
 		return nil, err
 	}
 	return stacks, nil
 }
 
+func (a *Adapter) UpdateStackLastTargetGroupARNs(stack *Stack) {
+	if _, ok := a.stackLastTargerGroupARNs[stack.Name]; !ok {
+		if len(stack.TargetGroupARNs) > 0 {
```
szuecs, line 640: why do you check this here?

aws/adapter.go
RomanZavodskikh, line 64: minor: Targer -> Target

aws/cf.go
```diff
@@ -500,13 +501,22 @@ func mapToManagedStack(stack *cloudformation.Stack) *Stack {
...
-func findManagedStacks(svc cloudformationiface.CloudFormationAPI, clusterID, controllerID string) ([]*Stack, error) {
+func findManagedStacks(svc cloudformationiface.CloudFormationAPI, clusterID, controllerID string, stacksLastTargetGroupARNs map[string][]string) ([]*Stack, error) {
 	stacks := make([]*Stack, 0)
 	err := svc.DescribeStacksPages(&cloudformation.DescribeStacksInput{},
 		func(page *cloudformation.DescribeStacksOutput, lastPage bool) bool {
 			for _, s := range page.Stacks {
 				if isManagedStack(s.Tags, clusterID, controllerID) {
-					stacks = append(stacks, mapToManagedStack(s))
+					stack := mapToManagedStack(s)
+					if len(stack.TargetGroupARNs) == 0 && stack.status == cloudformation.StackStatusRollbackInProgress {
```
szuecs, line 511: why check len(stack.TargetGroupARNs= == 0?

aws/adapter.go
```diff
@@ -248,6 +249,7 @@ func NewAdapter(clusterID, newControllerID, vpcID string, debug, disableInstrume
 		nlbCrossZone:               DefaultNLBCrossZone,
 		nlbHTTPEnabled:             DefaultNLBHTTPEnabled,
 		customFilter:               DefaultCustomFilter,
+		stackLastTargerGroupARNs:   make(map[string][]string),
```
RomanZavodskikh, line 252: minor: Targer -> Target

aws/cf.go
```diff
@@ -500,13 +501,22 @@ func mapToManagedStack(stack *cloudformation.Stack) *Stack {
...
+						if _, ok := stacksLastTargetGroupARNs[stack.Name]; ok {
+							log.Warnf("stack %s is in rolling back state, falling back to last saved output", stack.Name)
+							stack.TargetGroupARNs = stacksLastTargetGroupARNs[stack.Name]
+						} else {
+							log.Warnf("stack %s has no saved target groups, skipping", stack.Name)
```
szuecs, line 516: how do you skip here? I don't see that you skip anything :)

aws/adapter.go
MustafaSaber: The check can be moved outside the function yes, I just thought it made sense here.

aws/cf.go
MustafaSaber: I'm still thinking should it be for all states without output or only for rollback

aws/cf.go
MustafaSaber: Yes, thanks. Forgot the `continue` but I still need the code in worker.go that's where I reset and save "last" data

PR #679, opened 2024-02-01
Set the default ingress API Version

### LINKED ISSUE #678: There is no default ingressAPIVersion
With #515 the `ingressAPIVersion` variable is no longer set, because the cli flag was removed which specified a default. This means that when an adapter is created here: https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/master/controller.go#L377 then it is being passed an empty value.

The result we have seen in our systems is this that the requestURI being sent to the kubernetes API is `/apis//ingresses` which results in a 404. This basically tanks kube-ingress-aws-controller from the beginning.

jhuntwork:
Fixes #678

With the removal of the `--ingress-api-version` flag in 0.15.0, this value was no longer set and the constructed api URL is incomplete.


mikkeloscar:
:+1:

katyanna:
👍

jhuntwork:
Thanks for the quick review and merge!

PR #681, opened 2024-02-02
testsuit: add test case for flags

MustafaSaber:
Initial step to add tests to controller and get rid of kingpin, currently testing these flags isn't very flexible. Maybe we need to adopt something similar to skipper.


szuecs:
:+1:

MustafaSaber:
👍

PR #682, opened 2024-02-13
fix: disabling NLB cross zone load balancing had no effect on current NLBs

szuecs:
fix: disabling NLB cross zone load balancing had no effect on current NLBs

### DIFF
--- a/aws/cf_template.go
+++ b/aws/cf_template.go
@@ -300,11 +300,11 @@ func generateTemplate(spec *stackSpec) (string, error) {
 		)
 	}
 
-	if spec.nlbCrossZone && spec.loadbalancerType == LoadBalancerTypeNetwork {
+	if spec.loadbalancerType == LoadBalancerTypeNetwork {
 		lbAttrList = append(lbAttrList,
 			cloudformation.ElasticLoadBalancingV2LoadBalancerLoadBalancerAttribute{
 				Key:   cloudformation.String("load_balancing.cross_zone.enabled"),
-				Value: cloudformation.String("true"),
+				Value: cloudformation.String(fmt.Sprintf("%t", spec.nlbCrossZone)),
 			},
 		)
 	}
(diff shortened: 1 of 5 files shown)


szuecs:
:+1:

aws/cf_template.go
```diff
@@ -300,11 +300,11 @@ func generateTemplate(spec *stackSpec) (string, error) {
 		)
 	}
 
-	if spec.nlbCrossZone && spec.loadbalancerType == LoadBalancerTypeNetwork {
+	if spec.loadbalancerType == LoadBalancerTypeNetwork {
 		lbAttrList = append(lbAttrList,
 			cloudformation.ElasticLoadBalancingV2LoadBalancerLoadBalancerAttribute{
 				Key:   cloudformation.String("load_balancing.cross_zone.enabled"),
-				Value: cloudformation.String("true"),
+				Value: cloudformation.String(fmt.Sprintf("%t", spec.nlbCrossZone)),
```
AlexanderYastrebov, line 307: This allows disabling of crosszone for the existing NLBs created with cross-zone enabled because before the change this would just omit the attribute.
Is my understanding correct here?

szuecs: Yes

mikkeloscar: Are you sure you can change this on an existing NLB? I think it was left out because you can't change it after creation.

szuecs:
:+1:

AlexanderYastrebov:
:+1:

aws/cf_template.go
szuecs: Are you sure?
At least I can do this in the AWS console by clicking the edit button on attributes
image

szuecs: I tested it in our dev cluster and it works as expected. Maybe in the past it was not possible.

szuecs: Here random NLB from the dev cluster:
image

RomanZavodskikh:
:+1:

PR #694, opened 2024-04-30
Feature: nlb zone affinity via r53

szuecs:
feature: set nlb zone affinity for clients
log: log default cross zone and zone affinity config

This enables you to set dns_record.client_routing_policy configuration, such that you can decide for NLB via R53 feature that clients running in the same zone will hit also the same zone of your NLB. This is interesting for cross cluster traffic or traffic that stays in the same cluster but passes the NLB. Valid configurations are:
- availability_zone_affinity
- partial_availability_zone_affinity
- any_availability_zone (default)

see also https://docs.aws.amazon.com/elasticloadbalancing/latest/network/network-load-balancers.html\#load-balancer-attributes and https://docs.aws.amazon.com/elasticloadbalancing/latest/network/network-load-balancers.html\#zonal-dns-affinity

### DIFF
--- a/aws/adapter.go
+++ b/aws/adapter.go
@@ -66,6 +66,7 @@ type Adapter struct {
 	ipAddressType               string
 	albLogsS3Bucket             string
 	albLogsS3Prefix             string
+	nlbZoneAffinity             string
 	httpRedirectToHTTPS         bool
 	nlbCrossZone                bool
 	nlbHTTPEnabled              bool
@@ -124,6 +125,10 @@ const (
 	DefaultAlbS3LogsPrefix = ""
 
 	DefaultCustomFilter = ""
+
+	// DefaultZoneAffinity specifies dns_record.client_routing_policy, see also https://docs.aws.amazon.com/elasticloadbalancing/latest/network/network-load-balancers.html#load-balancer-attributes
+	DefaultZoneAffinity = "any_availability_zone"
+
 	// DefaultNLBCrossZone specifies the default configuration for cross
 	// zone load balancing: https://docs.aws.amazon.com/elasticloadbalancing/latest/network/network-load-balancers.html#load-balancer-attributes
 	DefaultNLBCrossZone   = false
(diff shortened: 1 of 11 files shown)


szuecs:
I want to test in a dev cluster before merge

aws/cf_template.go
```diff
@@ -307,6 +307,19 @@ func generateTemplate(spec *stackSpec) (string, error) {
 				Value: cloudformation.String(fmt.Sprintf("%t", spec.nlbCrossZone)),
 			},
 		)
+
+		println("HERE: ", spec.nlbZoneAffinity)
```
AlexanderYastrebov, line 311: HERE:

aws/adapter.go
```diff
@@ -124,6 +125,10 @@ const (
 	DefaultAlbS3LogsPrefix = ""
 
 	DefaultCustomFilter = ""
+
+	// DefaultZoneAffinity specifies dns_record.client_routing_policy, see also https://docs.aws.amazon.com/elasticloadbalancing/latest/network/network-load-balancers.html#load-balancer-attributes
+	DefaultZoneAffinity = "any_availability_zone"
```
AlexanderYastrebov, line 130: Maybe highlight that it is safe to change as per https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-elasticloadbalancingv2-loadbalancer-loadbalancerattribute.html#aws-properties-elasticloadbalancingv2-loadbalancer-loadbalancerattribute-properties

AlexanderYastrebov:
LGTM modulo minor comments.

AlexanderYastrebov:
:+1:

aws/cf_template.go
szuecs: oopsi :)

aws/adapter.go
szuecs: done

AlexanderYastrebov:
:+1:

szuecs:
:+1:

szuecs:
Default:
image

With zone affinity set and cross zone balancing off:
image

PR #699, opened 2024-05-07
containerization: use `amazonlinux-2023` base image for better security

MustafaSaber:
Signed-off-by: Mustafa Abdelrahman <mustafa.abdelrahman@zalando.de>

PR #704, opened 2024-06-03
deploy: use static base image

AlexanderYastrebov:
Use static base image which is similar to gcr.io/distroless/static

See previous #699


AlexanderYastrebov:
:+1:

Issue #723, opened 2024-10-09
AWS SDK v1  July 31, 2025 EOL


szuecs:
See https://github.com/aws/aws-sdk-go/issues/5270

> We previously announced the upcoming end-of-support for AWS SDK for Go (v1).
> 
> Per that announcement, as of 7/31/2024, the SDK has entered maintenance mode. Going forward, we will limit releases to address critical bug fixes and security issues only. The SDK will not receive API updates for new or existing services, or be updated to support new regions.
> 
> We are no longer accepting external contributions to this SDK.
> 
> Maintenance mode will last for 1 year. After 7/31/2025, the SDK will enter end-of-support, and no updates at all will be made. We recommend that you migrate to AWS SDK for Go v2. For additional details as well as information on how to migrate, please refer to the linked announcement.

we need to migrate to v2

PR #743, opened 2025-05-12
Update AWS Adapter APIs to use the controller context

ayruslore:
This change is next step after the AWS SDK Go v2 migration https://github.com/zalando-incubator/kube-ingress-aws-controller/issues/723, in this change we update the controller APIs to use the controller context and pass it to the AWS SDK.

fixes https://github.com/zalando-incubator/kubernetes-on-aws/issues/9396


certs/caching.go
```diff
@@ -87,11 +88,11 @@ func (cc *cachingProvider) updateCertCache() error {
 
 // startBackgroundRefresh creates a background forever loop to update
 // certificate cache.
-func (cc *cachingProvider) startBackgroundRefresh(certUpdateInterval time.Duration) {
+func (cc *cachingProvider) startBackgroundRefresh(ctx context.Context, certUpdateInterval time.Duration) {
 	go func() {
 		for {
 			time.Sleep(certUpdateInterval)
-			if err := cc.updateCertCache(); err != nil {
+			if err := cc.updateCertCache(ctx); err != nil {
```
AlexanderYastrebov, line 95: Here I think we can now exit loop when context is done and also use ticket instead of sleep like e.g. [here](https://github.com/zalando/skipper/blob/703d8e86f2c4b70cf9a93613052548584a53726b/scheduler/scheduler.go#L667-L678)

ayruslore: I have pushed a commit to update. Please check.

VERSION
```diff
@@ -1 +1 @@
-v0.16
+v17.0
```
AlexanderYastrebov, line 1: Should be v0.17

AlexanderYastrebov:
:+1:

PR #745, opened 2025-05-27
ingress-controller: update ingresses only if the load balancer is active

madumalt:
### Issue Description

As per the existing behavior, once the stack update completes, the
ingress controller immediately updates the status of all ingresses and
routegroups to reference the new load balancer. This update may
sometimes happens before the new load balancer has marked its targets
(skipper-ingress) as healthy, leading clients being routed to a load
balancer that is not ready to serve traffic yet.

### Fix

To improve this behavior we retrieve the ELBs via aws api and build
the models including the ELB state of the corresponding ELB. Then
before updating the ingresses/routegroups (updateIngress func) we
check whether the ELB is in active state, if not we skip updating
the ingresses/routegroups status with the ELB hostname.

### DIFF
--- a/aws/cf_template.go
+++ b/aws/cf_template.go
@@ -24,6 +24,13 @@ const (
 	//
 	// [0]: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticloadbalancingv2-listenerrule.html#cfn-elasticloadbalancingv2-listenerrule-priority
 	internalTrafficDenyRulePriority int64 = 1
+
+  // LoadBalancerResou  rceLogicalID is the logical ID of the LoadBalancer 
+  // resource in the CloudFormation template.
+  // 
+  // !!!Caution
+  //    Changing this value will recreate the LoadBalancer resource.
+	LoadBalancerResourceLogicalID = "LB"
 )
 
 func hashARNs(certARNs []string) []byte {
@@ -117,7 +124,7 @@ func generateTemplate(spec *stackSpec) (string, error) {
 	template.Outputs = map[string]*cloudformation.Output{
 		outputLoadBalancerDNSName: {
 			Description: "DNS name for the LoadBalancer",
-			Value:       cloudformation.GetAtt("LB", "DNSName").String(),
+			Value:       cloudformation.GetAtt(LoadBalancerResourceLogicalID, "DNSName").String(),
 		},
 		outputTargetGroupARN: {
 			Description: "The ARN of the TargetGroup",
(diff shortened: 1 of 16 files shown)


AlexanderYastrebov:
Overall approach to check loadbalancer status looks OK.
There should be a way to get loadbalancer ARN from CF stack by logical id (which is `"LB"`, see how CF stack is created).
DescribeLoadBalancers API allows fetching multiple load balancers so I think we should collect them first and then  buildManagedModel and processes updates.
Please also add tests.

worker.go
```diff
@@ -581,15 +582,25 @@ func isNoUpdatesToBePerformedError(err error) bool {
...
 }
 
-func updateIngress(kubeAdapter *kubernetes.Adapter, lb *loadBalancer, problems *problem.List) {
+func updateIngress(ctx context.Context, awsAdapter *aws.Adapter, kubeAdapter *kubernetes.Adapter, lb *loadBalancer, problems *problem.List) {
 	var dnsName string
 	if lb.clusterLocal {
 		dnsName = kubernetes.DefaultClusterLocalDomain
 	} else {
-		// only update ingress if the stack exists and is in a complete state.
+		// do not update ingress if the stack exists and is in a complete state.
```
AlexanderYastrebov, line 590: I think this was correct. If you want to negate the statement it should be "do not update ingress if the stack does not exist or is not in a complete state."

madumalt:
> so I think we should collect them first and then buildManagedModel and processes updates.

made the changes to address this

madumalt:
**TBD**

- [ ] TestResourceConversionOneToOne tests in worker_test.go is failing
- [ ] Add tests for the newly introduced funtions ()

aws/cf_template.go
```diff
@@ -24,6 +24,8 @@ const (
 	//
 	// [0]: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticloadbalancingv2-listenerrule.html#cfn-elasticloadbalancingv2-listenerrule-priority
 	internalTrafficDenyRulePriority int64 = 1
+
+	loadBalancerResourceLogicalID = "LB"
```
szuecs, line 28: add a comment that a change of this name will trigger either update or recreate (I am not sure) of all CF stacks owned by the controller in an AWS account.

madumalt:
Closing this PR in favour of https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/748. I messed the commit history while rebasing, it was easier to cherrypick the changes and create a new PR rather than trying to fix the commit history I messed up here.

PR #615, opened 2023-06-14
Move certificate mock to another package

MustafaSaber:
Follow up on https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/587

Separate certificate mock to another package.

PR #746, opened 2025-06-02
certs/fake: refactor CertificateProvider

AlexanderYastrebov:
Refactor fake certificate provider to enable creation of multiple certificates with the same CA and to allow manual mocking of GetCertificates result.

Follow up on #615 and #587

### DIFF
--- a/certs/fake/certificate.go
+++ b/certs/fake/certificate.go
@@ -8,100 +8,97 @@ import (
 	"crypto/x509/pkix"
 	"fmt"
 	"math/big"
-	"sync"
 	"time"
 
 	"github.com/zalando-incubator/kube-ingress-aws-controller/certs"
 )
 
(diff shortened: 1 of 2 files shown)


certs/fake/certificate.go
```diff
@@ -8,100 +8,97 @@ import (
...
-	once      sync.Once
-	err       error
+type CA struct {
 	chainKey  *rsa.PrivateKey
 	roots     *x509.CertPool
 	chainCert *x509.Certificate
 }
 
-type CertificateProvider struct {
-	ca caSingleton
```
AlexanderYastrebov, line 26: This was something wrong. I can understand singleton CA but when it is a member its a singleton within CertificateProvider instance and then it was only used once...

szuecs:
:+1:

PR #725, opened 2024-10-15
aws: fail on ListCertificatesPages error

AlexanderYastrebov:
`getACMCertificateSummaries` did not check `ListCertificatesPages` error
and returned empty `CertificateSummary` slice and no error
via `filterCertificatesByTag` if `filterTag` is configured.

Empty `CertificateSummary` results in empty `existingStackCertificateARNs`
for each `loadBalancer` model which triggered stack update:
```
level=info msg="Updating \"internet-facing\" stack for 0 certificates / 0 ingresses"
```
and then stack deletion on the next cycle:
```
level=info msg="Deleted orphaned stack \"kube-ingress-aws-controller-aws-XXX\""
```
due to empty certificate tags after previous update.

Follow up on https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/658

PR #747, opened 2025-06-02
worker: fail on missing certificates

AlexanderYastrebov:
Controller manages loadbalancer lifecycle based on available certificates.

Prevent controller from deleting loadbalancers when there are no certificates found e.g. due to a bug like #725 or a tag filter misconfiguration (#658).


AlexanderYastrebov:
:+1:

PR #748, opened 2025-06-16
ingress-controller: update ingresses only if the load balancer is active

madumalt:
As per the existing behavior, once the stack update completes, the ingress controller immediately updates the status of all ingresses and routegroups to reference the new load balancer. This update may sometimes happens before the new load balancer has marked its targets (skipper-ingress) as healthy, leading clients being routed to a load balancer that is not ready to serve traffic yet.

To improve this behavior we retrieve the ELBs via AWS API and build the models including the ELB states of the corresponding ELBs. Then before updating the ingresses/routegroups (updateIngress func) we check whether the ELB is in active state, if not we skip updating the ingresses/routegroups status with the ELB hostname.

### DIFF
--- a/aws/adapter.go
+++ b/aws/adapter.go
@@ -660,6 +660,42 @@ func (a *Adapter) FindManagedStacks(ctx context.Context) ([]*Stack, error) {
 	return stacks, nil
 }
 
+// GetStackLBStates returns the list of load balancers of the managed stacks.
+// While implementing this method it was taken that each stack has its own load
+// balancer and never refer to the same load balancer from multiple stacks.
+//
+// If a stack does not have a load balancer ARN it will be returned with a nil
+// LBState field.
+func (a *Adapter) GetStackLBStates(ctx context.Context, stacks []*Stack) ([]*StackLBState, error) {
+
+	lbARNs := make([]string, 0, len(stacks))
+	for _, stack := range stacks {
+		if stack.LoadBalancerARN == "" {
+			continue
+		}
+		lbARNs = append(lbARNs, stack.LoadBalancerARN)
+	}
+
+	lbsMap, err := getLoadBalancerStates(ctx, a.elbv2, lbARNs)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get load balancer states: %w", err)
+	}
+
+	stackLBStates := make([]*StackLBState, 0, len(stacks))
+	for _, stack := range stacks {
+		lbstate, found := lbsMap[stack.LoadBalancerARN]
+		if !found {
+			log.Warnf("The load balancer (ARN: %q) of %q stack is not found", stack.LoadBalancerARN, stack.Name)
+		}
+		stackELB := &StackLBState{
+			Stack:   stack,
+			LBState: lbstate,
+		}
+		stackLBStates = append(stackLBStates, stackELB)
+	}
+	return stackLBStates, nil
+}
+
 // UpdateTargetGroupsAndAutoScalingGroups updates Auto Scaling Groups
 // config to have relevant Target Groups and registers/deregisters single
 // instances (that do not belong to ASG) in relevant Target Groups.
--- a/aws/adapter_test.go
+++ b/aws/adapter_test.go
@@ -1028,3 +1028,138 @@ func TestWithTargetAccessMode(t *testing.T) {
 		assert.False(t, a.TargetCNI.Enabled)
 	})
 }
+
+func TestGetStackLBStates(t *testing.T) {
+	tests := []struct {
+		name                  string
+		elbv2Outputs          fake.ELBv2Outputs
+		stacks                []*Stack
+		expectedStackLBStates []*StackLBState
+		expectError           bool
+	}{
+		{
+			name: "success - matching ELBs",
+			elbv2Outputs: fake.ELBv2Outputs{
+				DescribeLoadBalancers: fake.R(&elbv2.DescribeLoadBalancersOutput{
+					LoadBalancers: []elbv2Types.LoadBalancer{
+						{
+							LoadBalancerArn: aws.String("arn1"),
+							State:           &elbv2Types.LoadBalancerState{Code: elbv2Types.LoadBalancerStateEnumActive},
+						},
+						{
+							LoadBalancerArn: aws.String("arn2"),
+							State:           &elbv2Types.LoadBalancerState{Code: elbv2Types.LoadBalancerStateEnumFailed},
+						},
+					},
+				}, nil),
+			},
+			stacks: []*Stack{
+				{Name: "stack1", LoadBalancerARN: "arn1"},
+				{Name: "stack2", LoadBalancerARN: "arn2"},
+			},
+			expectedStackLBStates: []*StackLBState{
+				{
+					Stack:   &Stack{Name: "stack1", LoadBalancerARN: "arn1"},
+					LBState: &LoadBalancerState{StateCode: elbv2Types.LoadBalancerStateEnumActive},
+				},
+				{
+					Stack:   &Stack{Name: "stack2", LoadBalancerARN: "arn2"},
+					LBState: &LoadBalancerState{StateCode: elbv2Types.LoadBalancerStateEnumFailed},
+				},
+			},
+			expectError: false,
+		},
+		{
+			name: "success - no matching ELB found for one of the stacks",
+			elbv2Outputs: fake.ELBv2Outputs{
+				DescribeLoadBalancers: fake.R(&elbv2.DescribeLoadBalancersOutput{
+					LoadBalancers: []elbv2Types.LoadBalancer{
+						{
+							LoadBalancerArn: aws.String("arn1"),
+							State:           &elbv2Types.LoadBalancerState{Code: elbv2Types.LoadBalancerStateEnumActive},
+						},
+					},
+				}, nil),
+			},
+			stacks: []*Stack{
+				{Name: "stack1", LoadBalancerARN: "arn1"},
+				{Name: "stack2", LoadBalancerARN: "arn2"},
+			},
+			expectedStackLBStates: []*StackLBState{
+				{
+					Stack:   &Stack{Name: "stack1", LoadBalancerARN: "arn1"},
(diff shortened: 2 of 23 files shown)


madumalt:
👍

madumalt:
👍

aws/adapter.go
```diff
@@ -660,6 +660,60 @@ func (a *Adapter) FindManagedStacks(ctx context.Context) ([]*Stack, error) {
...
+		}
+		lbARNs = append(lbARNs, stack.LoadBalancerARN)
+	}
+
+	lbsMap := make(map[string]*elbv2Types.LoadBalancer, len(lbARNs))
+
+	paginator := elbv2.NewDescribeLoadBalancersPaginator(
+		a.elbv2,
+		&elbv2.DescribeLoadBalancersInput{
+			LoadBalancerArns: lbARNs,
```
AlexanderYastrebov, line 684: What happens if we have more than 20 ARNs?

worker.go
```diff
@@ -288,10 +290,9 @@ func doWork(
 		return problems.Add("failed to list managed stacks: %w", err)
 	}
 
-	for _, stack := range stacks {
-		if err := stack.Err(); err != nil {
-			problems.Add("stack %s error: %w", stack.Name, err)
-		}
```
AlexanderYastrebov, line 294: I think we need to keep this loop to collect stack errors into problem list

worker.go
```diff
@@ -590,8 +592,24 @@ func updateIngress(kubeAdapter *kubernetes.Adapter, lb *loadBalancer, problems *
...
+		if !lb.stack.IsComplete() {
+			log.Infof(
+				"CF stack %q is not in a completed state, skipping ingress update",
+				lb.stack.Name,
+			)
+			return
+		}
+		if lb.state == nil || lb.state.Code != elbv2Types.LoadBalancerStateEnumActive {
+			log.Infof(
+				"the load balancer of CF stack %q is not in active state, skipping ingress update",
```
AlexanderYastrebov, line 610: Log messages should start from capital letter.
Lets log the state and maybe loadbalancer ARN as well?

aws/adapter.go
madumalt: Paginator takes care of it if I'm not mistaken. This is the way it was done with the stacks retrieval, I guess the same limitation applies there as well right?

madumalt: I'll double check this anyways

aws/adapter.go
```diff
@@ -660,6 +660,60 @@ func (a *Adapter) FindManagedStacks(ctx context.Context) ([]*Stack, error) {
...
+// While implementing this method it was taken that each stack has its own load
+// balancer and never refer to the same load balancer from multiple stacks.
+//
+// If a stack does not have a load balancer ARN it will be returned with an empty
+// ELB field.
+func (a *Adapter) GetStackELBs(ctx context.Context, stacks []*Stack) ([]*StackELB, error) {
+
+	lbARNs := make([]string, 0, len(stacks))
+	for _, stack := range stacks {
+		if stack == nil || stack.LoadBalancerARN == "" {
```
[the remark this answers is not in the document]
madumalt: since it is a pointer array it can happen, that's why add this check. It should not happen but if it happens, and if we do not have this then it'll be a panic

worker.go
madumalt: > Log messages should start from capital letter.

Interesintg! If I'm not mistaken in most of the common go libraries, they don't start logs with a capital letter.  I thought in go work logs should always start with a simple letter.

aws/cf.go
```diff
@@ -207,6 +213,7 @@ type denyResp struct {
 
 type CloudFormationAPI interface {
 	cloudformation.DescribeStacksAPIClient
+	cloudformation.ListStackResourcesAPIClient
```
madumalt, line 216: no longer needed

worker.go
madumalt: yeah true, will fix it

aws/adapter.go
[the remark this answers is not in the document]
madumalt: Well according [this one](https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/master/worker.go#L291-L295) even though there is a possibility we have been operating without a nill check so far. Will remove the nill check here as well.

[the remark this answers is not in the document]
madumalt: (removed nil check)

worker.go
madumalt: resolving this as the suggested changes were applied

aws/cf_template.go
```diff
@@ -24,6 +24,13 @@ const (
 	//
 	// [0]: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticloadbalancingv2-listenerrule.html#cfn-elasticloadbalancingv2-listenerrule-priority
 	internalTrafficDenyRulePriority int64 = 1
+
+	// LoadBalancerResou  rceLogicalID is the logical ID of the LoadBalancer
+	// resource in the CloudFormation template.
+	//
+	// !!!Caution
```
szuecs, line 31: we don't use such comments
Every comment should be only there if it's not obvious from code.
So I would be in favor of dropping this line

worker.go
```diff
@@ -25,6 +26,7 @@ type loadBalancer struct {
 	ingresses                    map[string][]*kubernetes.Ingress
 	scheme                       string
 	stack                        *aws.Stack
+	state                        *elbv2Types.LoadBalancerState
```
szuecs, line 29: worker should not no about AWS details only the business objects

aws/adapter.go
```diff
@@ -660,6 +660,60 @@ func (a *Adapter) FindManagedStacks(ctx context.Context) ([]*Stack, error) {
 	return stacks, nil
 }
 
+// GetStackELBs returns the list of load balancers of the managed stacks.
+// While implementing this method it was taken that each stack has its own load
+// balancer and never refer to the same load balancer from multiple stacks.
+//
+// If a stack does not have a load balancer ARN it will be returned with an empty
+// ELB field.
+func (a *Adapter) GetStackELBs(ctx context.Context, stacks []*Stack) ([]*StackELB, error) {
```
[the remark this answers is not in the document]
madumalt: > call the big function in "elbv2.go"

will do this change

aws/adapter.go
madumalt: [AWS API reference of DescribeLoadBalancers](https://docs.aws.amazon.com/elasticloadbalancing/latest/APIReference/API_DescribeLoadBalancers.html)

madumalt: Accommodated 20 loadBalancerARNs request param limit in DescribeLoadBalancers API with the [getLoadBalancerStates function](https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/748/files#diff-0efc4c5374da02ee834edf241846dcc4400c726b8b09af68804abef6a2f126d8R77-R119)

aws/adapter_test.go
```diff
@@ -1028,3 +1028,112 @@ func TestWithTargetAccessMode(t *testing.T) {
...
+		stacks            []*Stack
+		expectedStackELBs []*StackELB
+		expectError       bool
+	}{
+		{
+			name: "success - matching ELBs",
+			elbv2Outputs: fake.ELBv2Outputs{
+				DescribeLoadBalancers: fake.R(&elbv2.DescribeLoadBalancersOutput{
+					LoadBalancers: []elbv2Types.LoadBalancer{
+						{LoadBalancerArn: aws.String("arn1"), LoadBalancerName: aws.String("lb1")},
```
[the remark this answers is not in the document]
madumalt: covered in [here](https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/748/files#diff-211969a47cafc4a9cab22506ca9ce075711dc458b453c4f14d884c3d40a3150aR251-R284)

aws/cf_template.go
madumalt: removed the `!!!Caution` part and updated the comment

madumalt:
> there are changes in the architecture that we can not do, please check worker.go comment

this is why I like to go with a standard model like "hexagonal model" for project structuring, otherwise down the line it's difficult to maintain the initial structural intent of the project

aws/fake/elbv2.go
```diff
@@ -66,10 +70,45 @@ func (m *ELBv2Client) DescribeTargetGroups(context.Context, *elbv2.DescribeTarge
 	return out, m.Outputs.DescribeTargetGroups.err
 }
 
+func (m *ELBv2Client) DescribeLoadBalancers(ctx context.Context, input *elbv2.DescribeLoadBalancersInput, opts ...func(*elbv2.Options)) (*elbv2.DescribeLoadBalancersOutput, error) {
+	if input != nil && len(input.LoadBalancerArns) > 20 {
+		return nil, fmt.Errorf("too many load balancers requested: %d, maximum is 20", len(input.LoadBalancerArns))
```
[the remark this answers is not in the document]
madumalt: I tried doing that having it as a const, the issue here is, package `fake` is inside the package `aws`. The sensible place to have that constant is in the `aws` package. If we do so, it resulted in cyclic dependencies error.

The we should move the constant to another package, but I don't know whether it make sense. Then I thought since it is `fake` package that we use for unit tests it is fine to keep 20 hardcoded

worker.go
```diff
@@ -590,8 +598,28 @@ func updateIngress(kubeAdapter *kubernetes.Adapter, lb *loadBalancer, problems *
...
+			return
+		}
+		if !lb.stack.IsComplete() {
+			log.Infof(
+				"CF stack %q is not in a completed state, skipping ingress update",
+				lb.stack.Name,
+			)
+			return
+		}
+		if lb.state == nil || lb.state.StateCode != elbv2Types.LoadBalancerStateEnumActive {
```
szuecs, line 614: I think if you do this as a function in adapter, then you don't need to have elbv2Types in worker.go

aws/adapter.go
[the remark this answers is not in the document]
madumalt: moved the ELB retaled code segment to elbv2.go

madumalt:
👍

szuecs:
:+1:

PR #750, opened 2025-07-02
test: ensure the controller works on eks ipv6 clusters

ayruslore:
Test

Corresponding  PR https://github.com/zalando-incubator/kubernetes-on-aws/pull/9627/files

### DIFF
--- a/kubernetes/adapter.go
+++ b/kubernetes/adapter.go
@@ -183,9 +183,9 @@ func (a *Adapter) newIngress(typ IngressType, metadata kubeItemMetadata, host st
 		shared = false
 	}
 
-	ipAddressType := aws.IPAddressTypeIPV4
-	if getAnnotationsString(annotations, ingressALBIPAddressType, "") == aws.IPAddressTypeDualstack {
-		ipAddressType = aws.IPAddressTypeDualstack
+	ipAddressType := aws.IPAddressTypeDualstack
+	if getAnnotationsString(annotations, ingressALBIPAddressType, "") == aws.IPAddressTypeIPV4 {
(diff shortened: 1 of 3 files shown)


kubernetes/adapter.go
```diff
@@ -183,9 +183,9 @@ func (a *Adapter) newIngress(typ IngressType, metadata kubeItemMetadata, host st
 		shared = false
 	}
 
-	ipAddressType := aws.IPAddressTypeIPV4
-	if getAnnotationsString(annotations, ingressALBIPAddressType, "") == aws.IPAddressTypeDualstack {
-		ipAddressType = aws.IPAddressTypeDualstack
+	ipAddressType := aws.IPAddressTypeDualstack
```
AlexanderYastrebov, line 186: If you plan to promote this, lets make it backwards compatible by having a flag for the default address type.

kubernetes/adapter.go
```diff
@@ -227,8 +227,8 @@ func (a *Adapter) newIngress(typ IngressType, metadata kubeItemMetadata, host st
 	loadBalancerType = loadBalancerTypesIngressToAWS[loadBalancerType]
 
 	if loadBalancerType == aws.LoadBalancerTypeNetwork {
-		// ensure ipv4 for network load balancers
-		ipAddressType = aws.IPAddressTypeIPV4
+		// ensure both ipv4 and ipv6 for network load balancers
+		ipAddressType = aws.IPAddressTypeDualstack
```
AlexanderYastrebov, line 231: We need to understand why this special case was added and maybe we can remove it (or control with a flag) .

PR #753, opened 2025-07-21
worker: delay ingress status update for new load balancers

AlexanderYastrebov:
Within https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/748 the controller was improved to wait until loadbalancer state becomes "active" before updating ingress status which triggers traffic switch to the new loadbalancer (e.g. via external-dns).

It turns out that while new loadbalancer state is active its target groups may not have discovered targets yet.

This change delays ingress status update for a new loadbalancer to give its target groups time to discover targets. The delay is configured by a new `min-loadbalancer-age` flag. Default value was chosen after testing in an AWS account.

The more elaborate alternative is to call `describe-target-health` and check `TargetHealthDescriptions`.

Other changes:

* unexport loadbalancer state fields
* make loadbalancer state accessors its members
* skip logging of empty loadbalancer ARNs in GetStackLBStates
* use structured logging in updateIngress
* introduce AWS and Kubernetes API internal mocks using github.com/stretchr/testify/mock

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,13 @@ clean:
 ## test: runs go test
 test:
 	go test -v -race -coverprofile=profile.cov -cover $(GOPKGS)
-	grep -Ev 'github.com/zalando-incubator/kube-ingress-aws-controller/(certs/fake|aws/fake|internal/aws/cloudformation)/' profile.cov > profile.cov.tmp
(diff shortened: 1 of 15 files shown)


MustafaSaber:
👍

Issue #454, opened 2021-11-19
Reconsider CloudFormation template parameters usage


AlexanderYastrebov:
CloudFormation template to create load balancers was initially introduced within https://github.com/zalando-incubator/kube-ingress-aws-controller/issues/42

It seems the idea was to have a static parameterized template. Over time the template generation was translated into the go code #111 
> Changes cloudformation template from a yaml template to use the library https://github.com/mweagle/go-cloudformation to generate the template using Go. This is needed because we want to dynamically add certificates to the stack, which is not possible with pure cloudformation.
and became more and more conditional. 

Since template is in the code now it seems there is no value in having template parameters so we should consider removing them in favor of using values directly to simplify code structure (remove indirection), remove boilrerplate and ease maintenance.

AlexanderYastrebov:
Some of the parameters are used to get existing Stack configuration
https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/206b61372833c17d8aec21846a320286bd289fe2/aws/cf.go#L498-L502

I think these would need to be replaced e.g. with tags which makes little sense.

PR #757, opened 2025-07-23
aws: sort subnets in Adapter.FindLBSubnets

AlexanderYastrebov:
Adapter.FindLBSubnets method is used to specify stack subnets and becomes a parameter of a CloudFormation template.

This change returns sorted result to avoid sporadic changes to CloudFormation template that triggers update of CloudFormation stack and corresponding LoadBalancer Subnets.

Though LoadBalancer Subnets update does not require interruption as per https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-resource-elasticloadbalancingv2-loadbalancer.html#cfn-elasticloadbalancingv2-loadbalancer-subnets there is no reason to trigger any unneccessary updates.

Related to https://github.com/zalando-incubator/kube-ingress-aws-controller/issues/454

### DIFF
--- a/aws/adapter.go
+++ b/aws/adapter.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


ayruslore:
👍

PR #765, opened 2025-09-19
update actions to go 1.25

tcondeixa:
update actions to go 1.25


mikkeloscar:
> @mikkeloscar do we still use goveralls? This is failing everywhere and the project is not updated for some years.

I don't have a strong opinion on it. It's just my impression it normally "works" and this was a temporary error. But if there is something better then sure we can switch. I guess codacy is what we use internally.

tcondeixa:
👍

PR #785, opened 2026-03-24
packaging: add helm chart

MustafaSaber:
To make skipper more open source friendly and has an easy way to replace nginx-ingress, we need to make it easy for new teams to plug and play.

With AI making coding cheap in some cases, this is a good use case to actually maintain the helm chart and use AI to support with it.

The helm chart was created with https://github.com/zalando-incubator/kubernetes-on-aws/blob/dev/cluster/manifests/ingress-controller in consideration

To make skipper as an ingress controller package, we need:

- [ ] Make a package from this repo and add as an optional dependency in skipper package
- [ ] Use coreDNS & externalDNS also as optional dependencies
- [ ] Make/update skipper helm package

Todo:

- [ ] Test package in local cluster
- [ ] Test package in AWS cluster
- [ ] Test package in other cloud providers

### DIFF
--- a/.github/workflows/gh-packages.yaml
+++ b/.github/workflows/gh-packages.yaml
@@ -4,6 +4,10 @@ on:
     tags:
       - "*"
 
+permissions:
+  contents: read   # for actions/checkout
+  packages: write  # for pushing to GHCR (docker image + helm chart)
+
 env:
   REGISTRY: ghcr.io
   IMAGE_NAME: "${{ github.repository }}"
--- a/.github/workflows/gh-packages.yaml
+++ b/.github/workflows/gh-packages.yaml
@@ -66,3 +66,32 @@ jobs:
           push: ${{ github.event_name != 'pull_request' }}
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
+
+  helm-chart:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Extract chart version
+        id: chart-version
+        run: |
+          # Strip leading 'v' from the git tag (e.g. v0.20.1 -> 0.20.1)
+          echo "version=${GITHUB_REF_NAME#v}" >> "$GITHUB_OUTPUT"
+
+      - name: Login to GitHub Container Registry
+        run: |
+          helm registry login ghcr.io \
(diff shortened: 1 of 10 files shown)


szuecs:
> To make skipper as and ingress controller package, we need:
> 
>     * [ ]  Make a package from this repo and add as an optional dependency in skipper package
> 
>     * [ ]  Use coreDNS & externalDNS also as optional dependencies

Why CoreDNS?  
Why External-DNS should be optional?

> 
>     * [ ]  Make/update skipper helm package
> 
> 
> Todo:
> 
>     * [ ]  Test package in local cluster

Would be interesting how this will work. Maybe a "kind" implementation and maybe we need to use https://github.com/kubernetes-sigs/kindnet to figure out the network path and a local DNS target like CoreDNS that we could instrument with external-dns. 

> 
>     * [ ]  Test package in AWS cluster
> 
>     * [ ]  Test package in other cloud providers

I would say: just use AWS. It's the cloud provider with 80% customer share and the work and maintenance to ensure it works in other clouds with the changing cloud provider stack is too much work.

MustafaSaber:
> Why CoreDNS? Why External-DNS should be optional?

CoreDNS for east2west communication

ExternalDNS optional for people not using `kube-ingress-on-aws`  (or using one LB with one certificate)

> Would be interesting how this will work. Maybe a "kind" implementation and maybe we need to use https://github.com/kubernetes-sigs/kindnet to figure out the network path and a local DNS target like CoreDNS that we could instrument with external-dns.

I only tested this package locally (with `kind` cluster) and it was expected to fail

szuecs:
> > Why CoreDNS? Why External-DNS should be optional?
> 
> CoreDNS for east2west communication

That's maybe not necessary to have. It's more another feature and likely you do not want to break their DNS setup, so I would try to omit this.

> ExternalDNS optional for people not using `kube-ingress-on-aws` (or using one LB with one certificate)

But isn't the purpose of it to show a full working solution?
I would say skipper+kube-ingress-on-aws+ExternalDNS is mandatory. 

> 
> > Would be interesting how this will work. Maybe a "kind" implementation and maybe we need to use https://github.com/kubernetes-sigs/kindnet to figure out the network path and a local DNS target like CoreDNS that we could instrument with external-dns.
> 
> I only tested this package locally (with `kind` cluster) and it was expected to fail

But do you want to have it work?
We basically need only a service type load balancer pointing to skipper.

PR #788, opened 2026-04-20
worker: prevent SSL policy update loop on shared load balancers

MustafaSaber:
In https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/774, update LB in-place was introduced. However, it was also introducing the following bug

-> Ingress A has a custom SSL policy
-> Ingress B uses the default SSL policy
	-> Since Ingress B doesn't have the annotation we will allow it through as
        `(ingress.HasSSLPolicyAnnotation && l.sslPolicy != ingress.SSLPolicy)` doesn't apply
	-> So we add both ingresses to LB then later override LB SSL policy with new ingress SSL policy (default one)
Later `inSync` will revert to the old SSL policy and we go into infinite loop.

Adding the new tag will help keep in-place updates, while blocking mismatch ingresses from landing on the same LB

### DIFF
--- a/aws/adapter.go
+++ b/aws/adapter.go
@@ -812,7 +812,7 @@ func (a *Adapter) UpdateTargetGroupsAndAutoScalingGroups(ctx context.Context, st
 // All the required resources (listeners and target group) are created in a
 // transactional fashion.
 // Failure to create the stack causes it to be deleted automatically.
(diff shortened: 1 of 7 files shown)


madumalt:
👍

ayruslore (approved):

MustafaSaber:
fyi @lambertpandian

lambertpandian:
> fyi @lambertpandian

Thanks for the fix and thanks for updating me.

PR #789, opened 2026-04-22
Ingress: Mark ingresses with non-default SSL policy as non-shared

### LINKED ISSUE #790: Ingress LB recreation on cert timeout with custom SSL Policy
# Problem Overview

When a user creates one Ingress with SSL Policy annotation  and another without the annotation, at each `cert-ttl-timeout` the controller ended up creating a new LB and abandoning the existing one. 

It seems like this buggy behaviour introduced with https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/774

## How to reproduce.

1. Deploy and Ingress like the following,

```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  annotations:
    zalando.org/aws-load-balancer-ssl-policy: ELBSecurityPolicy-FS-2018-06
  labels:
    application: old-ssl-policy
  name: old-ssl-policy-1
  namespace: gift-cardl-lb-issue
spec:
  rules:
    - host: old-ssl-policy.thilina.exapmle.do
      http:
        paths:
          - backend:
              service:
                name: postgres-operator-ui
                port:
                  number: 80
            pathType: ImplementationSpecific
---
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  labels:
    application: old-ssl-policy
  name: old-ssl-policy-2
  namespace: gift-cardl-lb-issue
spec:
  rules:
    - host: old-ssl-policy.test.net
      http:
        paths:
          - backend:
              service:
                name: postgres-operator-ui
                port:
                  number: 80
            pathType: ImplementationSpecific
```

2. Wait until several `cert-ttl-timeouts` pass by
3. Filter controller logs for 'Deleted orphaned stack'

At each `cert-ttl-timeout` you should see this logline that indicates the deletion of an existing LB. Before you run the experiment make sure there are no case that outputs 'Deleted orphaned stack'. The complete log-line is as follows,

```
time="2026-04-21T13:56:47Z" level=info msg="Deleted orphaned stack \"kube-ingress-aws-controller-clustername-d0dc99a1-81fb-44ad-872f-1ca6333aefe2\""
```

madumalt:
Fixes: https://github.com/zalando-incubator/kube-ingress-aws-controller/issues/790

Remove HasSSLPolicyAnnotation field from Ingress struct and instead set Shared=false when a non-default SSL policy is annotated. This ensures ingresses with custom SSL policies get dedicated load balancers without needing a separate annotation check during LB matching.


mikkeloscar:
There is already #788 solving the problem in a different way.

> This ensures ingresses with custom SSL policies get dedicated load balancers without needing a separate annotation check during LB matching.

This implicitly mean that we need `n` LBs if we have `n` ingresses with a non-default SSL policy _even_ if the policy is the same for all of them. Before it would be merged to a single LB for that case.

madumalt:
> There is already #788 solving the problem in a different way.

~~Apparently it is not. I tested it in the pg9-euc1-test cluster and it does not fix the issue. It seems like it is not only about having Ingress with custom SSL Policy.~~

> > This ensures ingresses with custom SSL policies get dedicated load balancers without needing a separate annotation check during LB matching.
> 
> This implicitly mean that we need `n` LBs if we have `n` ingresses with a non-default SSL policy _even_ if the policy is the same for all of them. Before it would be merged to a single LB for that case.

Yeah this was an alternative solution and the solution is achieved by behaviour change. Tested it in the pg9-euc1-test cluster and it fixes the issue. 

You can see the comparison (tradeoffs) of these 2 PRs in [this doc](https://docs.google.com/document/d/1TtZXlhV2QznfZqaHTNfRM1AfWu05vXV4LZHq_AvyVa0) if interested.

madumalt:
closing in favour of https://github.com/zalando-incubator/kube-ingress-aws-controller/pull/788

Issue #794, opened 2026-05-04
opt-out annotation


szuecs:
I want to have an opt-out annotation, such that kube-ingress-aws-controller would ignore the ingress or routegroup if the annotation is set.
I propose to have `zalando.org/kube-ingress-aws-controller-opt-out` annotation so skipper will read the config but only this controller will not.
This is good for cases when you want to have routes but not a loadbalancer based on ingress or routegroup, but for example a service type loadbalancer. 
In our case the reason is to create an e2e test case for a canary skipper-ingress pod to test that this work as expected in e2e tests before it will be deployed to clusters.

AlexanderYastrebov:
Maybe `zalando.org/aws-load-balancer-type=none` ?

szuecs:
Good idea

PR #795, opened 2026-05-05
Add opt-out support for load-balancer-type: none

greeshma1196:
- [x] In `newIngress` (`kubernetes/adapter.go`): return `nil, nil` when `loadBalancerType == loadBalancerTypeNone`, before the NLB/WAF check and map lookup
- [x] In `ListIngress` and `ListRoutegroups`: guard append with `ing != nil`; log the exclusion at Info level
- [x] Add unit tests: Ingress and RouteGroup with `zalando.org/aws-load-balancer-type: none` are excluded from `ListResources`
- [x] Add `TestNewIngressFromRouteGroup` test case: RouteGroup with `zalando.org/aws-load-balancer-type: none` returns `nil, nil` from `newIngressFromRouteGroup` (parallel to the `TestNewIngressFromKube` case)
- [x] Add fixture entry with `zalando.org/aws-load-balancer-type: none` and class `skipper` to `testdata/fixture01.json` and `testdata/fixture01_rg.json`; assert count in `TestListIngress` stays at 1
- [x] Add fixture entry with `zalando.org/aws-load-balancer-type: none` and without class to testdata/fixture01.json and testdata/fixture01_rg.json; assert count in TestListIngress stays at 1
- [x] Update README
- [x] Bump minor version

Update:
- [x] Move the check for opt out out of the ingress business object
- [x] Add helper function `hasLoadBalancerTypeNone`
- [x] Remove `TestNewIngressFromRouteGroup` and the scenario in `TestNewIngressFromKube` - these were added because the previous logic existed within the ingress object, now we check for the load balancer type `none` early 

ref: https://github.com/zalando-incubator/kube-ingress-aws-controller/issues/794


kubernetes/adapter.go
```diff
@@ -54,6 +54,7 @@ const (
 	DefaultClusterLocalDomain = ".cluster.local"
 	loadBalancerTypeNLB       = "nlb"
 	loadBalancerTypeALB       = "alb"
+	loadBalancerTypeNone      = "none"
```
AlexanderYastrebov, line 57: It's worth checking the codebase for implicit assumptions about two loadbalancer types we might have e.g.:
```go
if typ == loadBalancerTypeNLB {
  ...
} else {
  // ALB assumed implicitly
}
```
I think reviewing all the usages of existing constants should confirm there are no assumptions.

AlexanderYastrebov: You'd also need to update the readme and maybe bump minor version.

greeshma1196: > I think reviewing all the usages of existing constants should confirm there are no assumptions.

Good point! Will check 😃

szuecs: hmm I would just drop the ingress early and filter it before we put this into the "ingress" business object as you do. Then all the assumptions are not important in the rest of the code.

kubernetes/testdata/fixture01.json
```diff
@@ -119,6 +119,28 @@
...
+      "metadata": {
+        "name": "fixture06",
+        "namespace": "default",
+        "selfLink": "/apis/networking.k8s.io/v1/namespaces/default/ingresses/fixture06",
+        "uid": "fixture06",
+        "resourceVersion": "42",
+        "generation": 1,
+        "creationTimestamp": "2016-11-29T14:53:42Z",
+        "annotations": {
+          "kubernetes.io/ingress.class": "skipper",
```
szuecs, line 133: another test without `kubernetes.io/ingress.class` would be good, so we make sure that we also do not add it if we run by default.

kubernetes/adapter.go
greeshma1196: > It's worth checking the codebase for implicit assumptions about two loadbalancer types we might have e.g.:
> 
> ```go
> if typ == loadBalancerTypeNLB {
>   ...
> } else {
>   // ALB assumed implicitly
> }
> ```
> 
> I think reviewing all the usages of existing constants should confirm there are no assumptions.

I went through the code, there are three places where we are making implicit assumptions:
https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/e7a053738c0c756cf152790fc8961a958b46ef27/aws/cf_template.go#L382
https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/e7a053738c0c756cf152790fc8961a958b46ef27/aws/cf_template.go#L481
https://github.com/zalando-incubator/kube-ingress-aws-controller/blob/e7a053738c0c756cf152790fc8961a958b46ef27/aws/adapter.go#L975

This does not affect our changes, because we ignore the ingress prior to hitting these sections.

greeshma1196: > hmm I would just drop the ingress early and filter it before we put this into the "ingress" business object as you do. Then all the assumptions are not important in the rest of the code.

I see your point, but I have a difference in opinion. Currently this is the only place where we read the annotations and set the load-balancer-type. If we have the check earlier, it will lead to duplication of code (small amount) elsewhere. Moreover the annotation exists in the ingress/routegroup, hence it should be a part of the ingress business logic. Please correct me if my thought process is incorrect.

kubernetes/testdata/fixture01.json
greeshma1196: added, please take a look.

kubernetes/adapter.go
greeshma1196: > You'd also need to update the readme and maybe bump minor version.

done, please take a look.

greeshma1196:
Test was conducted in a test cluster. 
Stackset used:
```
apiVersion: v1
kind: Namespace
metadata:
  name: gmapp1
---
apiVersion: zalando.org/v1
kind: StackSet
metadata:
  name: app1 
  namespace: gmapp1
  labels:
    application: app1
spec:
  ingress:
    metadata:
      annotations:
        zalando.org/aws-load-balancer-type: "none"
        zalando.org/skipper-routes: |
          r1: Path("/test") -> status(202) -> <shunt>;
    hosts:
    - test.pg9-test.zalan.do
    backendPort: 8080
  stackLifecycle:
    scaledownTTLSeconds: 60
    limit: 2
  stackTemplate:
    spec:
      version: v8
      replicas: 1
      podTemplate:
        spec:
          containers:
            - name: backend
              image: registry.opensource.zalan.do/teapot/skipper:latest
              ports:
                - containerPort: 8080
              args:
                - skipper
                - -address=:8080
                - -wait-for-healthcheck-interval=5s
              resources:
                limits:
                  cpu: 10m
                  memory: 50Mi
                requests:
                  cpu: 10m
                  memory: 50Mi
```

Loadbalancer was not provisioned, logs are attached:
```
time="2026-05-07T15:36:17Z" level=info msg="ingress resource is annotated with zalando.org/aws-load-balancer-type=none, skipping load balancer provisioning" name=app1-v8-traffic-segment ns=gmapp1 type=ingress
```

Also tested with zalando.org/aws-load-balancer-type=alb and without the annotations. The previous scenarios work as expected.

kubernetes/adapter.go
szuecs: A little duplication is here better than poluting the business object, imo.

szuecs (approved):

MustafaSaber (approved):

ayruslore (approved):

ayruslore (commented):
simple and clear. simply awesome. 🚀

